PR #4, opened 2024-04-23
Time provider moved from state manager to util package

Juliusan:
The code may be used to mock time in other packages, therefore it was extracted into separate test_util package. This was done on `iota2.0` branch and may be useful here too.

### DIFF
--- a/packages/util/time_util/interface.go
+++ b/packages/util/time_util/interface.go
@@ -1,0 +1,11 @@
+package time_util
+
+import (
+	"time"
+)
+
+type TimeProvider interface {
+	SetNow(time.Time)
+	GetNow() time.Time
+	After(time.Duration) <-chan time.Time
+}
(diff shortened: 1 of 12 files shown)


packages/util/time_util/time_provider_artifficial.go
```diff
@@ -1,4 +1,4 @@
-package sm_gpa_utils
```
jorgemmsilva, line 1: typo, should be `artificial`

packages/util/time_util/interface.go
```diff
@@ -0,0 +1,11 @@
+package time_util
+
+import (
+	"time"
+)
+
+type TimeProvider interface {
+	SetNow(time.Time)
```
jorgemmsilva, line 8: I know this exists already, but shouldn't `SetNow` be omitted from the "real" interface, only be present on the test implementation?

jorgemmsilva (commented):
just realized that I never clicked the "submit review" button :facepalm:

packages/util/time_util/interface.go
Juliusan: I do not remember the reasoning behind that. Probably I didn't want to complicate things too much.

PR #10, opened 2024-05-21
Initial address refactor

Juliusan:
Refactored Address interface of iota. New type is `cryptolib.Address`. Some other changes:
* Created `cryptolib.Signature` instead of `iotago.Signature`.
* Simplified `iotago.AddressSigner` interface into `cryptolib.Signer`, I hope this is not too simple.
* There are some functions which are temporary and we need to get rid off them later: `cryptolib.NewAddressFromIotago` and `cryptolib.Address.AsIotagoAddress` as well as `cryptolib.SignerToIotago` and `cryptolib.Signature.AsIotaSignature`.
Tried to make everything compile. Tried to solve as much as possible linter warnings, however avoided doing too much in order not to cause merge conflicts.


packages/isc/request_onledger.go
```diff
@@ -224,20 +225,20 @@ func (req *onLedgerRequestData) SenderAccount() AgentID {
 		return nil
 	}
 	if req.requestMetadata != nil && !req.requestMetadata.SenderContract.Empty() {
-		if sender.Type() == iotago.AddressAlias {
-			chainID := ChainIDFromAddress(sender.(*iotago.AliasAddress))
-			return req.requestMetadata.SenderContract.AgentID(chainID)
-		}
+		// if sender.Type() == iotago.AddressAlias {	// TODO: is it needed?
```
jorgemmsilva, line 228: it's not possible anymore to derive the owner type from an address. We cannot do these checks anymore unfortunately

PR #30, opened 2024-06-03
Isc types

lmoe:


packages/isc/sui/.bruno/SUI/bruno.json
```diff
@@ -0,0 +1,9 @@
+{
```
howjmay, line 1: are these bruno files added in purpose?

lmoe: Yes and no. 

The bruno files helped a lot testing the serialization. Right now it's just one route though. I work with multiple machines and that was the quickest way to have things in sync. :D 

Can eventually be removed.

howjmay: thanks understood

PR #41, opened 2024-06-12
Rename isc -> iscmove

dessaya:


sui-go/sui/api_impl.go
```diff
@@ -60,6 +60,6 @@ func NewSuiWebsocketClient(url string) *ImplSuiAPI {
 }
 
 const (
-	DefaultGasBudget uint64 = 10000000
+	DefaultGasBudget uint64 = 100000000
```
howjmay: I think we should change the gas budget in iscmove only

sui-go/iscmove/client.go
```diff
@@ -14,18 +14,19 @@ import (
...
 )
 
+// Client provides convenient methods to interact with the `isc` Move contracts.
 type Client struct {
-	// API *sui.ImplSuiAPI
 	*sui.ImplSuiAPI
 }
 
-func NewIscClient(api *sui.ImplSuiAPI) *Client {
+func NewClient(api *sui.ImplSuiAPI) *Client {
```
howjmay, line 22: I am curious about if we rename it to `NewClient()` isn't confusing with the sui client one?

dessaya: It's one of the suggestions here: https://go.dev/doc/effective_go#package-names

It's about avoiding repetition. The package name already contains `isc` so saying `isc.NewIscClient()` is repetitive. `isc.NewClient()` is more idiomatic.

sui-go/sui/api_impl.go
dessaya: Ok, I'll revert this change for now.

PR #51, opened 2024-06-17
Isc types

lmoe:
Updates the isc types
Refactor tests
Adds `state_index` to ISC contract
Adds GraphQL AssetBag query

Temporarily adds a mock contract for `packages/isc/sui` to test a filled AssetBag


sui-go/iscmove/client.go
```diff
@@ -35,18 +35,31 @@ func (c *Client) StartNewChain(
...
+	treasuryCap *models.SuiObjectResponse,
 ) (*models.SuiTransactionBlockResponse, error) {
 	ptb := sui_types.NewProgrammableTransactionBuilder()
-
 	// the return object is an Anchor object
+
+	arguments := []sui_types.Argument{}
+	if treasuryCap != nil {
+		ref := treasuryCap.Data.Ref()
+		f, _ := ptb.Obj(
```
howjmay: we can use MustObj() here

sui-go/iscmove/sui_graph/pull_schema.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+npx get-graphql-schema http://127.0.0.1:9001 > schema.graphql
```
howjmay, line 3: I have a lack of knowledge about graphQL. I am curious if this is querying the schema from a sui node?

lmoe: Yes exactly! I couldn't find the schema to download directly, so I did it manually. Guess we don't have to do that often, but this way it's at least documented.

sui-go/iscmove/client.go
lmoe: fixed

sui-go/iscmove/client.go
```diff
@@ -372,3 +386,7 @@ func (c *Client) ReceiveRequest(
 
 	return txnResponse, nil
 }
+
+func (c *Client) GetAssetBag(ctx context.Context, assetBagId *sui_types.ObjectID) {
+	c.ImplSuiAPI.GetDynamicFields(ctx, assetBagId, nil, nil)
```
howjmay: Is this aimed to not have return values?

lmoe: Nope, forgot to remove it. :D

packages/isc/sui/isc.go
```diff
@@ -36,9 +36,10 @@ type AssetBag struct {
...
 
 type Anchor struct {
-	ID        sui_types.ObjectID
-	Assets    AssetBag
-	StateRoot []uint8
+	ID         sui_types.ObjectID
+	Assets     Referent[AssetBag]
+	StateRoot  sui_types.Bytes
+	StateIndex uint32
 }
```
Juliusan, line 43: Is `Anchor` struct supposed to be mutable? I know, this is the Go way to do it, but I find it very unsafe to leave public fields in unmutable structs. Maybe switch to private fields and public getter methods? Of course, we can leave this change for later.

lmoe: That's a good point. They are partially mutable, the question is where. 

The StateIndex will increase. This can either happen on the contracts or ISCs side. The ID and the AssetsBag reference won't change.  

I would agree to add getter and make the fields private. Let me think about it a bit, considering there is the GraphQl implementation going.

sui-go/iscmove/sui_graph/pull_schema.sh
howjmay: thanks

PR #57, opened 2024-06-19
refactor: Make SuiParsedMoveObject.Fields into json.RawMessage

howjmay:


sui-go/models/objects.go
```diff
@@ -46,7 +47,7 @@ type SuiParsedMoveObject struct {
 	Type              string `json:"type"`
 	HasPublicTransfer bool   `json:"hasPublicTransfer"`
 	// maybe use reflect to solve this
```
dessaya: This comment is no longer necessary?

dessaya (approved):

sui-go/models/objects.go
howjmay: deleted

PR #70, opened 2024-06-25
fat: Impl simpler bigint struct

howjmay:


sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
+package utils
```
dessaya: Why not store it in package `models` like the rest of the objects in the library?

sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"math/big"
+	"strings"
+)
+
+type BigIntWrapper struct {
```
dessaya: I think it's safe (and better) to name it `BigInt`. No need for "wrapper" in the name.

sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
...
+	return &BigIntWrapper{new(big.Int).SetUint64(v)}
+}
+
+func (w *BigIntWrapper) UnmarshalText(data []byte) error {
+	return w.UnmarshalJSON(data)
+}
+
+func (w *BigIntWrapper) UnmarshalJSON(data []byte) error {
+	rawData := strings.TrimSpace(string(data))
+	if strings.HasPrefix(rawData, `"`) && strings.HasSuffix(rawData, `"`) {
```
dessaya: Why not something like:

```
var s string
json.Unmarshal(data, &s)
```

sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
...
+		rawData = rawData[1 : len(rawData)-1]
+	}
+	if w.Int == nil {
+		w.Int = new(big.Int)
+	}
+	if rawData == "null" {
```
dessaya: is it really necessary to accept `null` in json? Can't we just assume that it is always a string?

sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	_, ok := w.SetString(rawData, 10)
+	if ok {
+		return nil
+	}
+	return fmt.Errorf("json data [%s] is not T", string(data))
+}
+
+func (w *BigIntWrapper) MarshalJSON() ([]byte, error) {
+	if w.Int == nil {
```
dessaya: Is this check necessary? Can't we just assume that `w.Int` will never be nil?

sui-go/utils/bigint.go
```diff
@@ -0,0 +1,45 @@
...
+		return []byte(`"0"`), nil
+	}
+	return []byte(`"` + w.String() + `"`), nil
```
dessaya:
```
return json.Marshal(w.String())
```

sui-go/utils/bigint.go
howjmay: oh I was considering about this. I thought bigint is a more common thing that may share with other stuff, and `models` package aims for the object for API calls. 
If you think `models` is the better, then I can move it there. I am flexible for both places

sui-go/utils/bigint.go
howjmay: `data` could be integer too. I meant not `"123"` but `123`

sui-go/utils/bigint.go
howjmay: It is necessary. Some of the return values could be `null` from Sui nodes

sui-go/utils/bigint.go
dessaya: I think `models` is a good place for this type.

Even if we wanted to place it in another package, IMO `utils` is not a good name, since whenever someone puts something in a `utils` or `helpers` package it usually ends up containing all sorts of unrelated stuff.

sui-go/utils/bigint.go
dessaya: But the old `SafeSuiBigInt` did not have this ckeck... :thinking:

sui-go/utils/bigint.go
dessaya: Really? I mean, do we know for a fact that the Sui API encodes numbers as int in some places and string in others?

If some numbers are not supposed to be "big" then it is natural to encode them as int, but then we can just use `uint64` or whatever instead of `BigInt`.  `BigInt` should only be used for numbers that can be larger than 64 bits.

sui-go/utils/bigint.go
howjmay: I think `decimal` package may solve this. It basically totally relies on `decimal` package here

sui-go/utils/bigint.go
howjmay: Let me check. But I am pretty sure this issue happens, since I just triggered some error when I applied the above snippet

howjmay: Maybe I can leave a FIXME here, and fix it later?

dessaya: Ok

dessaya (approved):

PR #75, opened 2024-06-27
feat: Use request parameter for API calls

howjmay:


sui-go/utils/build.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/json"
+	"fmt"
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/iotaledger/wasp/sui-go/sui_types"
+)
+
+type CompiledMoveModules struct {
```
dessaya: You can use the `move.PackageBytecode` strut instead to avoid duplicating the definition

dessaya (approved):

PR #78, opened 2024-06-28
fix iscmove compiler errors (again)

dessaya:


clients/iscmove/client.go
```diff
@@ -42,7 +42,7 @@ func NewClient(config Config) *Client {
 	}
 }
 
-func (c *Client) RequestFunds(ctx context.Context, address cryptolib.Address) error {
+func (c *Client) RequestFunds(ctx context.Context, address *cryptolib.Address) error {
```
howjmay, line 42: Could you help me replace the implementation with this

```go
var faucetUrl string
	switch c.config.APIURL {
	case conn.TestnetEndpointUrl:
		faucetUrl = conn.TestnetFaucetUrl
	case conn.DevnetEndpointUrl:
		faucetUrl = conn.DevnetFaucetUrl
	case conn.LocalnetEndpointUrl:
		faucetUrl = conn.LocalnetFaucetUrl
	default:
		panic("not supported network")
	}
	err := sui.RequestFundFromFaucet(address.AsSuiAddress(), faucetUrl)
	if err != nil {
		return err
	}
	return nil
```

dessaya: No idea why this method ended up here. It seems to be a copy of `sui.RequestFundFromFaucet`. IMO we should remove one of these.

dessaya: Done

PR #81, opened 2024-07-03
Fix cryptolib, fix GetObjectInfo, fix tests

lmoe:
* Fixes cryptolib 
* Fixes tests
* Fixes client not using Address correctly
* Fixes `GetCreatedObjectInfo` to use Created items again
* Made `KeySchemeFlagDefault = KeySchemeFlagIotaEd25519`

(This will make ISC incompatible with official SUI testnets, but is the only way to use our iota-test-validator)

### DIFF
--- a/packages/cryptolib/compatability_test.go
+++ b/packages/cryptolib/compatability_test.go
@@ -1,0 +1,33 @@
+package cryptolib
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/iotaledger/wasp/sui-go/sui"
+	"github.com/iotaledger/wasp/sui-go/suisigner"
+)
+
+const (
+	SEEDFORUSER = "0x1234"
+)
+
+func TestCompatability(t *testing.T) {
+	seedIndex := uint32(1)
+	seed := SEEDFORUSER
+
+	kp := KeyPairFromSeed(SubSeed(sui.MustAddressFromHex(seed)[:], seedIndex))
+	subseed := SubSeed(sui.MustAddressFromHex(seed)[:], seedIndex)
+	suikp := suisigner.NewSigner(subseed[:], suisigner.KeySchemeFlagEd25519)
+
+	require.Equal(t, kp.Address().AsSuiAddress().Data(), suikp.Address().Data())
+
+	kpSign, err := kp.SignTransactionBlock([]byte{1, 2, 3, 4}, suisigner.DefaultIntent())
+	require.NoError(t, err)
+
+	suiSign, err := suikp.SignTransactionBlock([]byte{1, 2, 3, 4}, suisigner.DefaultIntent())
+	require.NoError(t, err)
+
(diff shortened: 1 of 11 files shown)


packages/cryptolib/compatability_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	SEEDFORUSER = "0x1234"
+)
+
+func TestCompatability(t *testing.T) {
+	seedIndex := uint32(1)
+	seed := SEEDFORUSER
+
+	kp := KeyPairFromSeed(SubSeed(sui.MustAddressFromHex(seed)[:], seedIndex))
```
howjmay: Do you mind to use 
```go
seed := make([]byte, 32)
rand.Read(token)
```
To replace the `SEEDFORUSER` which is a string?

packages/cryptolib/compatability_test.go
```diff
@@ -0,0 +1,33 @@
...
+	subseed := SubSeed(sui.MustAddressFromHex(seed)[:], seedIndex)
+	suikp := suisigner.NewSigner(subseed[:], suisigner.KeySchemeFlagEd25519)
```
howjmay: Maybe we should use `KeySchemeFlagIotaEd25519`? 
I think we may eliminate the usage of other KeySchmes. After I don't think we will support them in near future

packages/cryptolib/public_key.go
```diff
@@ -9,9 +9,10 @@ import (
 	"go.dedis.ch/kyber/v3"
 	"go.dedis.ch/kyber/v3/group/edwards25519"
 
+	"github.com/minio/blake2b-simd"
```
howjmay, line 12: Do we use different blake2b package deliberately?

sui-go/suisigner/key_schemes.go
```diff
@@ -7,7 +7,7 @@ import (
 
 type KeySchemeFlag byte
 
-var KeySchemeFlagDefault = KeySchemeFlagEd25519
+var KeySchemeFlagDefault = KeySchemeFlagIotaEd25519
```
howjmay, line 10: This may break the tests in suisigner

howjmay: But we can ignore them temporarily. Not super important

packages/cryptolib/public_key.go
lmoe: Previously we only needed the blake2b library for the legacy derivation of the seed, but with Sui it is needed for hashing the public key. I just used the same blake library that was used in the legacy derivation. No specific choice on my side here.

sui-go/suisigner/key_schemes.go
lmoe: Fixed it partially. ExampleSigner breaks because the mnemonic wont fit the keyscheme, but the others work fine :)

howjmay (approved):

PR #82, opened 2024-07-03
feat: Finish ISC contracts

howjmay:


clients/iscmove/client_anchor.go
```diff
@@ -0,0 +1,231 @@
...
+					TypeArguments: []sui.TypeTag{},
+					Arguments:     []sui.Argument{argAnchor, argReqObject},
+				},
+			},
+		)
+		ptb.Command(
+			sui.Command{
+				TransferObjects: &sui.ProgrammableTransferObjects{
+					Objects: []sui.Argument{
+						{NestedResult: &sui.NestedResult{Cmd: uint16(i * 2), Result: 1}},
```
jorgemmsilva, line 119: what does this do exactly?

If I understand it's collecting the assetbags, but we should take the assets out of those bags:
https://github.com/iotaledger/isc-private/blob/develop/sui-go/contracts/isc/tests/anchor_tests.move#L99-L121

PR #85, opened 2024-07-04
Client and metadata refactoring  

lmoe:
* Move L1/L2 client into `/clients`
* Generated interface for sui client (as L1Client)
* Embedded suiclient into our own "Extended" L1Client for additional methods (Health, RequestFunds)
* Added Token metadata as a constant value instead of L1() params 
* Brings back Bech32 
* Refactored Cluster environment
* Makes wasp and wasp-cli compilable

### DIFF
--- a/clients/l1client.go
+++ b/clients/l1client.go
@@ -16,42 +16,6 @@ type L1Config struct {
 	APIURL    string
 }
 
-type L1ClientExt struct {
-	*suiclient.Client
-
-	Config L1Config
-}
-
-func (c *L1ClientExt) RequestFunds(ctx context.Context, address cryptolib.Address) error {
-	var faucetURL string = c.Config.FaucetURL
-	if faucetURL == "" {
-		switch c.Config.APIURL {
-		case suiconn.TestnetEndpointURL:
-			faucetURL = suiconn.TestnetFaucetURL
-		case suiconn.DevnetEndpointURL:
-			faucetURL = suiconn.DevnetFaucetURL
-		case suiconn.LocalnetEndpointURL:
-			faucetURL = suiconn.LocalnetFaucetURL
-		default:
-			panic("unspecified FaucetURL")
-		}
-	}
-
-	return suiclient.RequestFundsFromFaucet(ctx, address.AsSuiAddress(), faucetURL)
-}
-
-func (c *L1ClientExt) Health(ctx context.Context) error {
-	_, err := c.Client.GetLatestSuiSystemState(ctx)
-	return err
-}
-
-func NewL1Client(l1Config L1Config) L1Client {
-	return &L1ClientExt{
-		suiclient.New(l1Config.APIURL),
-		l1Config,
-	}
-}
-
 type L1Client interface {
 	GetDynamicFieldObject(
 		ctx context.Context,
@@ -233,4 +197,43 @@ type L1Client interface {
 	RequestFunds(ctx context.Context, address cryptolib.Address) error
 	Health(ctx context.Context) error
 	WithWebsocket(url string)
+}
+
+var _ L1Client = &L1ClientExt{}
+
+type L1ClientExt struct {
+	*suiclient.Client
+
+	Config L1Config
+}
+
+func (c *L1ClientExt) RequestFunds(ctx context.Context, address cryptolib.Address) error {
+	faucetURL := c.Config.FaucetURL
+
+	if faucetURL == "" {
+		switch c.Config.APIURL {
+		case suiconn.TestnetEndpointURL:
+			faucetURL = suiconn.TestnetFaucetURL
+		case suiconn.DevnetEndpointURL:
+			faucetURL = suiconn.DevnetFaucetURL
+		case suiconn.LocalnetEndpointURL:
+			faucetURL = suiconn.LocalnetFaucetURL
+		default:
+			panic("unspecified FaucetURL")
+		}
+	}
+
+	return suiclient.RequestFundsFromFaucet(ctx, address.AsSuiAddress(), faucetURL)
+}
+
+func (c *L1ClientExt) Health(ctx context.Context) error {
+	_, err := c.Client.GetLatestSuiSystemState(ctx)
+	return err
+}
+
+func NewL1Client(l1Config L1Config) L1Client {
+	return &L1ClientExt{
+		suiclient.New(l1Config.APIURL),
+		l1Config,
+	}
 }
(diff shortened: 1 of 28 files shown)


clients/l1client.go
```diff
@@ -0,0 +1,239 @@
...
+		ctx context.Context,
+		req suiclient.TryMultiGetPastObjectsRequest,
+	) ([]*suijsonrpc.SuiPastObjectResponse, error)
+	WithSignerAndFund(seed []byte, index int) (*suiclient.Client, suisigner.Signer)
+	RequestFunds(ctx context.Context, address cryptolib.Address) error
+	Health(ctx context.Context) error
+	WithWebsocket(url string)
+}
+
+var _ L1Client = &L1ClientExt{}
```
howjmay, line 202: Not a problem. Just my personal question. What does Ext stand for?

lmoe: Extended :D 
Wasn't really sure about the name but just went with it

packages/nodeconn/nc_chain.go
```diff
@@ -293,17 +293,21 @@ func (ncc *ncChain) postTxLoop(ctx context.Context) {
...
-		ctxWithTimeout, ctxCancel := context.WithTimeout(nodeConn.ctx, inxTimeoutBlockMetadata)
-		defer ctxCancel()
-
-		// check if the transaction was already included (race condition with other validators)
-		if _, err := ncc.nodeConn.nodeClient.TransactionIncludedBlock(ctxWithTimeout, pendingTx.ID(), parameters.L1().Protocol); err == nil {
-			// transaction was already included
-			pendingTx.SetConfirmed()
-			return true
-		}
+		panic("refactor me: parameters.L1()")
```
howjmay, line 296: Not sure what do you think about it. I would use FIXME (in capital) to hint the places to I need add something or fix which I can use some extension to easily find. 
Not sure whether using a uniform pattern can help us locate all the places need to be enhance. 
But it is totally fine to keep the current ones

packages/parameters/l1parameters.go
```diff
@@ -17,13 +18,16 @@ type L1Params struct {
 	BaseToken      *BaseToken                 `json:"baseToken" swagger:"required"`
 }
 
+const CoinTypeBaseToken suijsonrpc.CoinType = "0x2::sui::SUI"
```
howjmay: we have a constant `suijsonrpc.SuiCoinType` which is `0x2::sui::SUI`. Maybe use it here?

packages/parameters/l1parameters.go
```diff
@@ -17,13 +18,16 @@ type L1Params struct {
...
+
 type BaseToken struct {
-	Name            string `json:"name" swagger:"desc(The base token name),required"`
-	TickerSymbol    string `json:"tickerSymbol" swagger:"desc(The ticker symbol),required"`
-	Unit            string `json:"unit" swagger:"desc(The token unit),required"`
-	Subunit         string `json:"subunit" swagger:"desc(The token subunit),required"`
-	Decimals        uint32 `json:"decimals" swagger:"desc(The token decimals),required"`
-	UseMetricPrefix bool   `json:"useMetricPrefix" swagger:"desc(Whether or not the token uses a metric prefix),required"`
+	Name            string              `json:"name" swagger:"desc(The base token name),required"`
```
howjmay, line 22: Normally I use `sui.Identifier` for coin name in sui-go. It is just an alias to show it is a module name or object name in sui. Not sure if we should use it as datatype of Name here

packages/parameters/l1parameters.go
lmoe: It will change to `0x2::iota::IOTA` eventually I guess, so it rather belongs in ISC space than in the sui-go lib imo

tools/cluster/cluster.go
```diff
@@ -826,37 +829,26 @@ func (clu *Cluster) ActiveNodes() []int {
...
-}
-
 func (clu *Cluster) AddressBalances(addr *cryptolib.Address) *isc.Assets {
 	// get funds controlled by addr
-	outputMap, err := clu.l1.OutputMap(addr)
+
+	balances, err := clu.l1.GetAllBalances(context.Background(), addr.AsSuiAddress())
 	if err != nil {
-		fmt.Printf("[cluster] GetConfirmedOutputs error: %v\n", err)
+		log.Panicf("[cluster] failed to GetAllBalances for address[%v]", addr.Bech32(parameters.Bech32Hrp))
```
howjmay: Should we use logger in hive.go?

packages/parameters/l1parameters.go
lmoe: Yeah maybe. 
I'm not sure if we want to keep the structure at all, but it is required by other modules right now. So I just made it available.

packages/nodeconn/nc_chain.go
lmoe: I think the time is up to change to a FIXME right now, I already removed way too much in previous PRs :D 

But I kept the "refactor me:" pattern for every related change

clients/l1client.go
howjmay: ah ook!

packages/parameters/l1parameters.go
howjmay: But I think `suijsonrpc.SuiCoinType` will be renamed to IOTA too. It would be so strange that we host something with Sui naming

tools/cluster/cluster.go
lmoe: Good point, it should be `clu.log` instead.

packages/parameters/l1parameters.go
howjmay: okok! The sui.Identifier is only alias, it behaves 100% the same as string. It is more like a notification than a real syntax behavior

packages/nodeconn/nc_chain.go
howjmay: ooook! I was afraid of we may forget these someday. But it should be quite hard tbh. Sorry being annoyingly picky

packages/parameters/l1parameters.go
lmoe: You are totally right, it makes no sense otherwise. I'll change it. :D

howjmay (approved):

packages/nodeconn/nc_chain.go
lmoe: Picky is good :)

PR #88, opened 2024-07-11
add RequestsFeed client (wip)

dessaya:
Adds a new client `iscmove.RequestsFeed` that replaces the old `l1connection`.

It is a work in progress, for now it has 2 methods:
    * `SubscribeOwnedRequests` that fetches currently owned `Request` objects
    * `SubscribeNewRequests` that subscribes to newly received `Request` objects via websocket

Also a few minor refactors:

* created constants for module and object names like `"anchor"` and `"Anchor"`
* removed duplicate code regarding the `FaucetURL`
* split `suiclient.Client` and `suiclient.WebsocketClient` (since you can only call `Subscribe*` on a websocket client)
* added `suiclient.UnmarshalBCS()` as a shortcut for `bcs.Unmarshal()` that also verifies that the consumed bytes is exactly len(data)


jorgemmsilva (approved):

sui-go/suijsonrpc/objects.go
```diff
@@ -183,18 +183,13 @@ type SuiObjectResponse struct {
 
 type CheckpointSequenceNumber = uint64
 
-type CheckpointedObjectID struct {
```
howjmay, line 186: @dessaya May I ask the reason to delete this?

dessaya: Reading the docs of `getOwnedObjects`: https://docs.sui.io/sui-api-ref#suix_getownedobjects

I saw that the `cursor` parameter is just a `ObjectID`. So maybe this `CheckpointedObjectID` corresponds to a previous version of the API?

howjmay: ah you are right. 
I saw something interesting here, `suix_queryObjects` exists in the rust node codebase, but doesn't exist in the doc. cool 
But thank you

PR #102, opened 2024-07-18
Vm refactor

lmoe:
Too much to summarize. 

Refactors the VM to support SUI change


clients/iscmove/client_anchor.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/fardream/go-bcs/bcs"
 
+	"github.com/iotaledger/wasp/clients/iscmove/iscmove_types"
```
dessaya: Using `_` in names is unconventional.

Anyway, maybe I would leave the types in `iscmove`, and move the client to `iscmoveclient` (with no `_`).

Or maybe it's time to move those types to `isc`.

packages/isc/request.go
```diff
@@ -108,6 +92,7 @@ func MustLogRequestsInTransaction(tx *iotago.Transaction, log func(msg string, a
 	}
 }
 
+// TODO: Refactor me:
 // RequestsInTransaction parses the transaction and extracts those outputs which are interpreted as a request to a chain
 func RequestsInTransaction(tx *iotago.Transaction) (map[ChainID][]Request, error) {
```
dessaya, line 97: I think this function should go away. Requests are no longer included in transactions, but rather sent as objects to the anchor.

packages/isc/sandbox_interface.go
```diff
@@ -120,7 +121,7 @@ type Sandbox interface {
 	Send(metadata RequestParameters)
 	// EstimateRequiredStorageDeposit returns the amount of base tokens needed to cover for a given request's storage deposit
 	EstimateRequiredStorageDeposit(r RequestParameters) uint64
-	// StateAnchor properties of the anchor output
+	// StateAnchor properties of the anchor request
```
dessaya, line 124: hmmm if you did a global replace "output" -> "request" I think it will have made some incorrect changes like here. It's correct for "request outputs" but not "anchor output" or "NativeToken output", "NFT output"...

packages/vm/core/accounts/internal.go
```diff
@@ -82,13 +82,13 @@ const (
 	// Covered in: TestDepositNFTWithMinStorageDeposit
 	keyNFTOwner = "NW"
 
-	// keyNewNativeTokens stores an array of <nativeTokenID>, containing the newly created native tokens that need filling out the OutputID
+	// keyNewNativeTokens stores an array of <nativeTokenID>, containing the newly created native tokens that need filling out the RequestID
```
dessaya: Another example where the global replace went wrong :(

packages/vm/gas/table.go
```diff
@@ -73,15 +76,9 @@ const (
...
-	g := constGas
-	return func(_ uint64) uint64 {
-		return g
-	}
-}
-
-func (c BurnCode) Cost(p ...uint64) uint64 {
-	x := uint64(0)
+// Cost computes the cost based on the burn code and parameters
+func (c BurnCode) Cost(p ...*big.Int) *big.Int {
```
dessaya: Wait... I think these functions operate mostly on "gas units", not "gas fees".

I think it's ok to keep gas units as u64. Gas fees should move to bigInt because they are an amount of tokens, but gas units is a different thing.

sui-go/contracts/isc/sources/request.move
```diff
@@ -23,6 +23,8 @@ module isc::request {
     /// Represents a request object
     public struct Request has key {
         id: UID,
+        /// The anchor address
+        anchor: address,
```
dessaya: Why is this needed?

lmoe: It points to the anchor the request is related to. It's used as the "TargetAddress" property in the VM.

packages/isc/request.go
lmoe: Yes this is true. I spent half a day to refactor (remove) this function, but it's too often used in other parts. It has 8 direct references, but those functions are referenced multiple times again and requires a new logic. 

I think this rather deserves it's own PR.

sui-go/contracts/isc/sources/request.move
dessaya: But the Request move object is already owned by the anchor. Isn't this duplicate information?

lmoe: This depends on the angle you are looking at it. The VM only gets a list of Requests and they do not include the Anchor Address in a readily available form. We would need to either enrich the Request inside isc itself or the Request needs to provide it itself. Thought that's the simplest way. But open for suggestions! :)

dessaya: But the VM should already know the anchor address because it is a parameter stored in the VMTask object. (previously VMTask.AnchorOutput)
The anchor address can also be derived from the ChainID? I don't remember how we define the ChainID now.

lmoe: The VM should know yes, the only question for me was, how important it is to be able to fully recreate the OnLedgerRequest soley off an iscmove.Request data packet. If we include the Anchor address in the Request, this is possible. If not, then not.

Not sure if it's needed or not, but seemed important which is why I added it. We can remove it, and artificially add the anchor address in ISC.

dessaya: I don't see the need for it but I may be wrong. @jorgemmsilva wdyt?

jorgemmsilva: I agree with @dessaya . The VM should know the chain it is executing the requests for. No need to have this in each individual request as well. 

We had this in the past because it was how requests outputs were constructed. It shouldn't be necessary for iota-rebased ISC.

PR #108, opened 2024-07-25
feat: Add AssetsBagWithBalancesToAssets

howjmay:


clients/iscmove/client_assets_bag.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/fardream/go-bcs/bcs"
 
 	"github.com/iotaledger/wasp/packages/cryptolib"
+	"github.com/iotaledger/wasp/packages/isc"
```
lmoe, line 11: This won't work. The import will cause an import cycle. 

I think `AssetsBagWithBalancesToAssets` should rather be part of isc.Assets and therefore be called `ISCAssetsFromAssetsBag` or something. 

isc should point to iscmove, not the other way round.

lmoe (changes requested):

lmoe:
Minor fix required, but overall looks good :)

clients/iscmove/client_assets_bag.go
howjmay: I found putting this in isc.Assets will cause cyclic importing. That is why I put it here. And the test passed in my local side. I thought it should be ok

lmoe: That is a problem. I tried it this way and it caused me import cycles during build. It seems like we get import cycles no matter if its inside isc.Assets or in iscmove. Meh.

lmoe: I think we need to move the iscmove types into a separate namespace. 

I need the iscmove types in the isc and this is whats causing the import cycle. 

Lets just merge it, I'll fix it.

lmoe (approved):

clients/iscmove/client_assets_bag.go
howjmay: May I ask how can I trigger the cyclic importing error you mentioned?

Maybe we can move isc.Assets under iscmove?

howjmay: Thanks

PR #153, opened 2024-08-01
refactor: define interface for txbuilder

jorgemmsilva:


howjmay (approved):

packages/vm/vmtxbuilder/interface.go
```diff
@@ -0,0 +1,23 @@
...
+	ConsumeRequest(req isc.OnLedgerRequest) (storageDepositNeeded *big.Int)
+	SendObject(object sui.Object) (storageDepositReturned *big.Int)
+
+	// TODO there a "dry-run" mode, but that is "l1 server-side". Is there away to verifyt his without an external request?
+	MustBalanced()
+
+	// TODO
+	// - add stateMetadata?
+	// - what exactly needs to be signed by the committee
+	BuildTransactionEssence(stateRoot *state.L1Commitment) (msg []byte)
```
dessaya: My guess is `sui.TransactionData`, which is an unsigned tx that can later be marshalled with bcs and signed with `signer.SignTransactionBlock()`

dessaya (approved):

packages/vm/vmtxbuilder/interface.go
howjmay: agree with this

howjmay: Or `sui.ProgrammableTransaction` in this way we don't need to assign gas object in txbuilder

packages/vm/vmtxbuilder/interface.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/iotaledger/wasp/sui-go/sui"
+)
+
+type TransactionBuilder interface {
+	Clone() TransactionBuilder
+	ConsumeRequest(req isc.OnLedgerRequest) (storageDepositNeeded *big.Int)
+	SendObject(object sui.Object) (storageDepositReturned *big.Int)
+
+	// TODO there a "dry-run" mode, but that is "l1 server-side". Is there away to verifyt his without an external request?
+	MustBalanced()
```
howjmay: I am curious about one thing.
The total asset that vm can process is `(all request's assets) + (assets owned by anchor)` 
What is the possible scenario that about input out put maybe unbalanced?

every time updating an Anchor object there could be multiple input (Request objects) about should be only one output, because all the request object will be destroyed and only anchor object will be mutated?

jorgemmsilva: yeah I also think that won't be possible. This is more of a "tx is valid" sanity check kinda. Let's delete it for now :+1:

PR #167, opened 2024-08-08
fix suijsonrpc.Balance

howjmay:


sui-go/suijsonrpc/bigint.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 )
 
+type Uint128 = BigInt
```
lmoe, line 10: Do we even need any kind of BigInt reference here anymore? Is it used for other things than coin values?

clients/iscmove/iscmoveclient/client_assets_bag.go
```diff
@@ -225,7 +225,7 @@ func (c *Client) GetAssetsBagWithBalances(
 		cointype := suijsonrpc.CoinType("0x" + data.Name.Value.(string))
 		bag.Balances[cointype] = &suijsonrpc.Balance{
 			CoinType:     cointype,
-			TotalBalance: moveBalance.Value,
+			TotalBalance: moveBalance.Value.Uint64(),
```
lmoe, line 228: Shouldn't moveBalance.Value not already be of type uint64 here?

lmoe (approved):

sui-go/suijsonrpc/balance.go
```diff
@@ -7,10 +7,10 @@ import (
 type CoinType = string
 
 type Balance struct {
-	CoinType        CoinType          `json:"coinType"`
-	CoinObjectCount uint64            `json:"coinObjectCount"`
-	TotalBalance    *BigInt           `json:"totalBalance"`
-	LockedBalance   map[BigInt]BigInt `json:"lockedBalance"` // FIXME the type may not be wrong
+	CoinType        CoinType            `json:"coinType"`
+	CoinObjectCount uint64              `json:"coinObjectCount"`
+	TotalBalance    uint64              `json:"totalBalance"`
```
dessaya, line 12: Doesn't the jsonrpc api return the balance in string form? e.g. `"0x1a2b3c4d"`

sui-go/suijsonrpc/bigint.go
howjmay: This is for the LockedBalance which in the rust impl is in Uint128. So I think using BigInt should be ok here

clients/iscmove/iscmoveclient/client_assets_bag.go
howjmay: I think it is a string in the json response that is why I unmarshal it into bigint first then convert it to uint64

sui-go/suijsonrpc/balance.go
howjmay: yeah, but I think it is easier to handle the scalar number, so I convert it in the place above that Lukas commented

clients/iscmove/iscmoveclient/client_assets_bag.go
lmoe: But why into BigInt? All coin values are or will be changed to uint64 (same for LockedBalance, I guess?)

dessaya: I think it's ok to use BigInt in the suijsonrpc, because the BigInt type is tailored to marshal json in the same format that the jsonrpc expects.

Maybe the name is misleading, should be something like `RPCInt` or such.

PR #168, opened 2024-08-08
chore: refactor typings

lmoe:


packages/isc/assets.go
```diff
@@ -104,6 +104,14 @@ func (c CoinBalances) Add(coinType coin.Type, amount coin.Value) CoinBalances {
 	return c
 }
 
+func (c CoinBalances) Set(coinType coin.Type, amount coin.Value) CoinBalances {
+	if amount == 0 {
+		return c
```
dessaya, line 110: should probably be `delete c[coinType]`

packages/vm/core/evm/evmimpl/iscmagic_sandbox.go
```diff
@@ -120,28 +121,28 @@ func (h *magicContractHandler) Send(
...
-	if req.Assets.BaseTokens < sd {
+	if uint64(req.Assets.BaseTokens()) < sd {
 		if !req.AdjustToMinimumStorageDeposit {
 			panic(errBaseTokensNotEnoughForStorageDeposit.Create(
 				req.Assets.BaseTokens,
 				sd,
 			))
 		}
-		req.Assets.BaseTokens = sd
+		req.Assets.SetBaseTokens(coin.Value(sd))
```
dessaya: The cast should not be necessary because `EstimateRequiredStorageDeposit` should return `coin.Value`

dessaya (approved):

PR #172, opened 2024-08-12
test: Add test for txbuilder

howjmay:


packages/cryptolib/address.go
```diff
@@ -84,6 +84,11 @@ func (a *Address) AsSuiAddress() *sui.Address {
 	return &result
 }
 
+func FromSuiAddress(a *sui.Address) *Address {
```
dessaya: There is already a `NewAddressFromSui` in `isc-private` branch

dessaya (approved):

packages/cryptolib/address.go
howjmay: ummm but when will we merge it tho

dessaya: it's already merged

PR #175, opened 2024-08-13
refactor: Update start_new_chain()'s func params

howjmay:


packages/apilib/deploychain.go
```diff
@@ -47,7 +47,18 @@ func DeployChain(ctx context.Context, par CreateChainParams, stateControllerAddr
 		originatorAddr, stateControllerAddr, par.N, par.T)
 	fmt.Fprint(textout, par.Prefix)
 
-	anchor, err := par.Layer1Client.L2().StartNewChain(ctx, par.OriginatorKeyPair, par.PackageID, nil, suiclient.DefaultGasPrice, suiclient.DefaultGasBudget, par.InitParams.Bytes(), false)
+	anchor, err := par.Layer1Client.L2().StartNewChain(
+		ctx,
+		par.OriginatorKeyPair,
+		par.PackageID,
+		par.InitParams.Bytes(),
+		nil, // FIXME add this field
```
howjmay: @dessaya do you know what should I use at here

dessaya: The result of `origin.InitChain()` or so.

If you can't figure it out don't worry, there may still be some missing pieces.

howjmay: How about my current way adding blockHash and stateRoot in `CreateChainParams`?

dessaya: It's probably ok, and if not we can always change it later

howjmay: great

dessaya (approved):

PR #178, opened 2024-08-15
some progress with evm core contract

dessaya:


packages/evm/evmtest/ISCTest.sol
```diff
@@ -49,60 +48,79 @@ contract ISCTest {
...
+        require(allowance.coins[0].amount > TokensForGas);
+        assets.coins[0].amount = allowance.coins[0].amount - TokensForGas;
 
         ISCSendMetadata memory metadata;
         ISCSendOptions memory options;
         ISC.sandbox.send(receiver, assets, true, metadata, options);
     }
 
-    function sendNFT(L1Address memory receiver, NFTID id, uint64 storageDeposit) public {
+    function sendNFT(SuiAddress receiver, SuiObjectID id, uint64 storageDeposit) public {
```
lmoe, line 74: While the name SuiAddress is technically correct, I would rather name it IotaAddress as we will internally have to change all SuiXYZ names to IotaXYZ rather sooner than later.

lmoe (approved):

packages/evm/evmtest/ISCTest.sol
dessaya: Yeah I get that. Personally I prefer postponing the "Iota" rename, since right now we are sui compatible. I can do the rename when we decide to.

lmoe: With the change of the uint64/coin.Value we are not sui compatible anymore, as the balance values are big.Int in Sui

dessaya: I'm pretty sure values are u64 in Sui as well.
See for example the Balance struct in https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/docs/sui-framework/balance.md#0x2_balance_Balance

PR #180, opened 2024-08-17
feat: Add RotationTransaction

howjmay:


clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -181,3 +181,18 @@ func NewReceiveRequestPTB(
 
 	return ptb.Finish(), nil
 }
+
+func NewRotationTransaction(anchorRef *sui.ObjectRef, rotationAddress *sui.Address) (sui.ProgrammableTransaction, error) {
```
kape1395: It sounds like it, but it should return TransactionData, not just ProgrammableTransaction.
Maybe I'm confused about it.

howjmay: May I ask where and how this func will be called? I think gas fee and gas object are accessible only with connecting to L1. However, if we call this in txbuilder, then it would cause some problems, because txbuilder doesn't have access to L1 I think

kape1395: Currently, it is called here: https://github.com/iotaledger/isc-private/blob/bd73241b59b5e69f6ef5616ce080648f15d26d90/packages/chain/cons/cons.go#L588

Maybe I have to supply some context here. All the nodes have to supply the same coin for the gas.
They probably have to agree on other fields, too, to build the same TX on all the nodes.
What are the parameters, and how can I get them?

howjmay: I think we should pass a context to build the rotate transaction which contains gas object and gas price and maybe gas budget. 
But we should ensure not to reuse the same ObjectRef. Updating ObjectRef should be necessary if we have a ObjectRef in the function context
I guess we should use nodeconn to access these parameters. Not sure how @dessaya design these

dessaya: I don't think I designed anything around this :stuck_out_tongue: 

But I agree that returning `TransactionData` is probably more accurate here. Anything that is needed to construct the transaction (sender, gas fee, etc) can be received as a parameter.

clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -181,3 +181,18 @@ func NewReceiveRequestPTB(
...
+	ptb := sui.NewProgrammableTransactionBuilder()
+
+	ptb.Command(
```
kape1395: By the way, wouldn't it be better to keep all the logic in the move code and call the corresponding move functions in the PTB?

howjmay: Do you mean the logic of `RotationTransaction`? It is only a `TransferObjects` func call, so `NewRotationTransaction` construct the targeting move func.

clients/iscmove/iscmoveclient/client_anchor_ptb.go
howjmay: I am curious where should I receive the parameters we need here? And I think there should be a component keep updating the owning gas object and latest gas price?

howjmay:
I think the current way to call this 
` rotationTX, err := vmtxbuilder.NewRotationTransaction(vmResult.RotationAddress.AsSuiAddress()) `
will cause errors. We need to ensure every time a `vm.VMTask` is executed, a Ref of gas object will be consumed. However, rotate the ownership of anchor object will create another transaction which needs to consume another Ref of gas object. However, we haven’t update the Ref here. Therefore the Ref will be used multiple times which will panic.

clients/iscmove/iscmoveclient/client.go
```diff
@@ -66,7 +66,7 @@ func (c *Client) SignAndExecutePTB(
 ) (*iotajsonrpc.IotaTransactionBlockResponse, error) {
 	signer := cryptolib.SignerToIotaSigner(cryptolibSigner)
 	var err error
-	if len(gasPayments) == 0 {
+	if len(gasPayments) == 0 || gasPayments[0] == nil {
```
dessaya: This is suspicious. Is it expected that the slice might contain nil values?

packages/vm/vmtxbuilder/txbuilder.go
```diff
@@ -107,12 +109,19 @@ func (txb *AnchorTransactionBuilder) SendCrossChainRequest(targetPackage *iotago
...
 func (txb *AnchorTransactionBuilder) BuildTransactionEssence(stateMetadata []byte, topUpAmount uint64) iotago.ProgrammableTransaction {
 	if txb.ptb == nil {
 		txb.ptb = iotago.NewProgrammableTransactionBuilder()
 	}
 	// we have to discard the current txb to avoid reusing an ObjectRef
-	defer func() { txb.ptb = nil }()
+	defer func() {
+		txb.ptb = nil
+		txb.rotateToAddr = nil
+	}()
```
dessaya: This should not be necessary, as the AnchorTransactionBuilder is used only once by the VM.

dessaya (approved):

packages/vm/vmtxbuilder/txbuilder.go
howjmay: removed

clients/iscmove/iscmoveclient/client.go
howjmay: This was trying to avoid some case in the test. I just removed it.

PR #184, opened 2024-08-23
Reflection-based BCS serialiazation

nnikolash:


packages/isc/agentid_contract.go
```diff
@@ -10,8 +10,8 @@ import (
 
 // ContractAgentID is an AgentID formed by a ChainID and a contract Hname.
 type ContractAgentID struct {
-	chainID ChainID
-	hname   Hname
+	chainID ChainID `bcs:""`
```
dessaya, line 13: What is the purpose of `bcs:""`?

packages/util/bcs/options.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	return nil
+}
+
+func FieldOptionsFromTag(a string) (_ FieldOptions, _ error) {
+	if a == "" {
+		return FieldOptions{}, nil
+	}
+	if a == "-" {
```
dessaya, line 52: Maybe calling it `"skip"` or `"omit"` instead of `"-"` is more readable?

dessaya (approved):

packages/isc/agentid_contract.go
nnikolash: This is needed to mark unexported field for serialization.

By default only public fields are serialized, while private fields only serialized if tag is present. I've decided to introduce this logic, because throught existing code there are lot of places where we serialize structs with just private fields, so I found this functionality useful to avoid having creating special DTO type every time. But I was also afraid that engineeers will acidentally serialize something they did not intended, so I added this restriction.

I would immediatelly agree that the logic and style of this feature are arguable, so I would love to hear more opinions and consider other options.

packages/util/bcs/options.go
nnikolash: I would choose `"-"` because it is rather standard style - standard JSON and YAML serializers have it, mapstructure have it, and many others.
We may also add `"skip"` or `"omit"` as alternative to `"-"`, but that might be confusing

PR #189, opened 2024-08-27
Use new types in solo package (incomplete)

vitaliy-io:


packages/vm/core/accounts/interface.go
```diff
@@ -80,4 +80,7 @@ var (
 	ViewTotalAssets = coreutil.NewViewEP01(Contract, "totalAssets",
 		coreutil.FieldWithCodec(codec.CoinBalances),
 	)
+
+	// TODO: implement. This is a mock just to get stuff compiling
+	FuncNativeTokenCreate = coreutil.NewEP0(Contract, "funcNativeTokenCreate")
```
dessaya: I removed this because we will not implement minting coins for now. We can remove or comment out any calls from solo as well.

vitaliy-io: Ok got it. I added it just to make some code compiling. Will remove

dessaya (approved):

PR #193, opened 2024-08-30
Use new types in solo 2

vitaliy-io:


packages/solo/ledgerl1l2.go
```diff
@@ -71,80 +72,91 @@ func (ch *Chain) L2Assets(agentID isc.AgentID) *isc.Assets {
 func (ch *Chain) L2AssetsAtStateIndex(agentID isc.AgentID, stateIndex uint32) *isc.Assets {
```
dessaya: This should return `CoinBalances` instead of `Assets` IMO. As a general rule, if a function only deals with coins and not NFTs/objects, it should deal with `CoinBalances`.

packages/solo/ledgerl1l2.go
```diff
@@ -29,15 +30,15 @@ func (ch *Chain) L2Accounts() []isc.AgentID {
...
-}
+// func (ch *Chain) parseAccountBalance(d dict.Dict, err error) *isc.Assets {
+// 	require.NoError(ch.Env.T, err)
+// 	if d.IsEmpty() {
+// 		return isc.NewEmptyAssets()
+// 	}
+// 	ret, err := isc.AssetsFromDict(d)
+// 	require.NoError(ch.Env.T, err)
+// 	return ret
+// }
```
dessaya: It's ok to delete this

packages/solo/ledgerl1l2.go
```diff
@@ -71,80 +72,91 @@ func (ch *Chain) L2Assets(agentID isc.AgentID) *isc.Assets {
...
+// 	return out.(*iotago.FoundryOutput), nil
+// }
+
+// func (ch *Chain) GetNativeTokenIDByFoundrySN(sn uint32) (iotago.NativeTokenID, error) {
+// 	o, err := ch.GetFoundryOutput(sn)
+// 	if err != nil {
+// 		return iotago.NativeTokenID{}, err
+// 	}
+// 	return o.MustNativeTokenID(), nil
+// }
```
dessaya: ok to delete

packages/vm/core/accounts/state.go
```diff
@@ -48,7 +48,7 @@ func NewStateWriterFromSandbox(ctx isc.Sandbox) *StateWriter {
 }
 
 // converts an account key from the accounts contract (shortform without chainID) to an AgentID
-func agentIDFromKey(key kv.Key, chainID isc.ChainID) (isc.AgentID, error) {
+func AgentIDFromKey(key kv.Key, chainID isc.ChainID) (isc.AgentID, error) {
```
dessaya, line 51: Why this change?

dessaya (approved):

packages/solo/ledgerl1l2.go
vitaliy-io: done

packages/solo/ledgerl1l2.go
vitaliy-io: done

packages/vm/core/accounts/state.go
vitaliy-io: I needed to create AgentID in L2Accounts() func in ledgerl1l2.go

packages/solo/ledgerl1l2.go
vitaliy-io: done

PR #205, opened 2024-09-03
State manager unit tests pass

Juliusan:
Fixes to make state manager tests pass.

I am not sure about 
```
-	"math/rand/v2"
+	"math/rand"
```
but I needed that to make tests compile. Is my Golang version too old?


dessaya (approved):

howjmay (approved):
maybe we should migrate to math/rand/v2 someday? It is a new feature from 1.22
I am still reading it too
https://go.dev/blog/randv2

PR #213, opened 2024-09-09
refactor sandbox

howjmay:


howjmay:
Send cross chain request to the anchor which belongs to another isc-package is still under test

sui-go/contracts/isc/sources/anchor.move
```diff
@@ -65,6 +65,22 @@ module isc::anchor {
...
+        borrow::put_back(&mut anchor.assets, assets, b);
+    }
+
+    public fun take_and_place_to_assets_bag<T>(anchor: &mut Anchor, assets_bag: &mut AssetsBag, amount: u64) {
+        let (mut assets, b) = borrow::borrow(&mut anchor.assets);
+        let bal = assets.take_coin_balance<T>(amount);
+        assets_bag.place_coin_balance(bal);
+        borrow::put_back(&mut anchor.assets, assets, b);
+    }
+
```
dessaya: Are these functions really needed? If they are just utility functions, can't they be implemented in Go? I think it's better to keep the Move code as small as possible, because it is harder to upgrade whenever we need changes.

sui-go/contracts/isc/sources/assets_bag.move
```diff
@@ -52,6 +52,12 @@ module isc::assets_bag {
         place_coin_balance_internal(self, balance)
     }
 
+    public fun place_coin_with_amount<T>(self: &mut AssetsBag, coin: &mut Coin<T>, amount: u64, ctx: &mut TxContext) {
+        let split_coin = coin.split(amount, ctx);
+        let balance = coin::into_balance(split_coin);
+        place_coin_balance_internal(self, balance)
+    }
```
dessaya: Same here, can't this be implemented in Go?

sui-go/contracts/isc/sources/anchor.move
howjmay: I will try to make them into PTB

howjmay: fixed

dessaya (approved):

PR #215, opened 2024-09-11
refactor: Make PTB construction composable

howjmay:


clients/iscmove/iscmoveclient/client_allowance.go
```diff
@@ -26,7 +26,9 @@ func (c *Client) AllowanceNew(
 	var err error
 	signer := cryptolib.SignerToSuiSigner(cryptolibSigner)
 
-	ptb := NewAllowanceNewPTB(packageID, cryptolibSigner.Address())
+	ptb := sui.NewProgrammableTransactionBuilder()
+	ptb = NewAllowanceNewPTB(ptb, packageID, cryptolibSigner.Address())
```
dessaya: This is better! But the function names are now confusing IMO. This one should be `PTBNewAllowance` or `NewAllowancePTB` or `PTBAllowanceNew` or such. Because it is not creating a "new PTB".

howjmay: I think I will remove all the prefix `New` in all the PTB funcs. The 2nd New here means it is the `new()` move func in the `allowance` module

dessaya (approved):

PR #233, opened 2024-09-17
Remove blob

cwarnerdev:


tools/wasp-cli/chain/deploycontract.go
```diff
@@ -36,7 +34,6 @@ func initDeployContractCmd() *cobra.Command {
 			client := cliclients.WaspClient(node)
 			vmtype := args[0]
 			name := args[1]
-			description := args[2]
 			initParams := util.EncodeParams(args[4:], chainID)
```
howjmay: If we removed the description field, we may need to change the index of args here too?
```diff
- initParams := util.EncodeParams(args[4:], chainID)
+ initParams := util.EncodeParams(args[3:], chainID)
```

tools/wasp-cli/chain/deploycontract.go
```diff
@@ -51,13 +48,8 @@ func initDeployContractCmd() *cobra.Command {
 				log.Check(err)
```
howjmay, line 48: And 
```diff
- progHash, err = hashing.HashValueFromHex(args[3])
+ progHash, err = hashing.HashValueFromHex(args[2])
```

howjmay (changes requested):

packages/origin/origin.go
```diff
@@ -83,15 +82,13 @@ func InitChain(
...
-		blob.Contract,
 		accounts.Contract,
 		blocklog.Contract,
 		errors.Contract,
 		governance.Contract,
 		evm.Contract,
 	})
-	blob.NewStateWriter(blob.Contract.StateSubrealm(d)).SetInitialState()
-	accounts.NewStateWriter(v, accounts.Contract.StateSubrealm(d)).SetInitialState(originDeposit, baseTokenCoinInfo)
+	accounts.NewStateWriter(v, accounts.Contract.StateSubrealm(d)).SetInitialState(originDeposit)
```
dessaya: `baseTokenCoinInfo` is needed here

packages/vm/execution/common.go
```diff
@@ -14,7 +15,7 @@ func GetProgramBinary(ctx WaspContext, programHash hashing.HashValue) (vmtype st
 	if ok {
 		return vmtype, nil, nil
 	}
-	return ctx.LocateProgram(programHash)
+	return "", nil, fmt.Errorf("native processor not found for programHash")
```
dessaya: This should probably be one of `coreerrors`

dessaya (approved):

packages/origin/origin.go
cwarnerdev: oops, messed that up with a merge conflict resolution. thanks!

PR #236, opened 2024-09-17
store SuiCoinInfo in accounts core contract

dessaya:


packages/isc/suicoininfo.go
```diff
@@ -0,0 +1,42 @@
+package isc
+
+import (
+	"github.com/iotaledger/wasp/packages/coin"
+	"github.com/iotaledger/wasp/packages/util/bcs"
+	"github.com/iotaledger/wasp/sui-go/suijsonrpc"
+)
+
+type SuiCoinInfo struct {
```
howjmay, line 9: Not sure if this would make confusion between the Sui coin? Maybe call it as `CoinMetadata` or `CoinInfo` directly?

packages/vm/core/accounts/basetokens.go
```diff
@@ -64,18 +64,18 @@ func (s *StateReader) GetBaseTokensBalanceDiscardExtraDecimals(agentID isc.Agent
 	return bts
 }
 
-func remainderKey(accountKey kv.Key) kv.Key {
-	return prefixRemainders + accountKey
+func accountWeiRemainderKey(accountKey kv.Key) kv.Key {
```
howjmay, line 67: The smallest unit of Sui is Mist. Should we rename Wei to Mist or the basic unit of IOTA? Not really necessary, just curious about it.

packages/isc/suicoininfo.go
dessaya: How is it confusing?

From other packages it is referenced as `isc.SuiCoinInfo`. I think it's a good thing to have the Sui prefix to emphasize that it contains information about a Sui coin.

packages/vm/core/accounts/basetokens.go
dessaya: I agree that "wei" is confusing here. But I picked it because we store the smallest unit in Ethereum (which has 18 decimals), so it is definitely not "mist".

PR #240, opened 2024-09-18
refactor allowance interface

howjmay:


clients/iscmove/isc.go
```diff
@@ -114,13 +106,18 @@ type Message struct {
 	Args     [][]byte
 }
 
+type AllowanceElement struct {
```
dessaya: I'd call it `AllowanceItem` but that's just preference.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -22,7 +23,8 @@ func (c *Client) CreateAndSendRequest(
 	iscContractHname uint32,
 	iscFunctionHname uint32,
 	args [][]byte,
-	allowanceRef *sui.ObjectRef,
+	allowanceCoinTypes []suijsonrpc.CoinType,
+	allowanceBalances []uint64,
```
dessaya: Can't this be a single array of `AllowanceElement` / `AllowanceItem` ?

clients/iscmove/iscmoveclient/client_request_ptb.go
```diff
@@ -13,7 +14,9 @@ func PTBCreateAndSendRequest(
 	iscContractHname uint32,
 	iscFunctionHname uint32,
 	args [][]byte,
-	argAllowance sui.Argument,
+	// the order of the allowance will be reversed after processed by move
```
dessaya: Why is the order reversed?

dessaya (approved):

clients/iscmove/isc.go
dessaya: or `CoinAllowance` (because in the future we should add NFTs / objects)

clients/iscmove/iscmoveclient/client_request_ptb.go
howjmay: In the move contract, I am using `pop_back()` to take elements from the input vector. 
If we want the order to be the same then we need to use `remove()`, however, this will cause more gas.
I think the order should not be a prominent value, so I decided to use `pop_back()`

clients/iscmove/iscmoveclient/client_request.go
howjmay: I can do it in the golang side, but the interface of the move function should be the same as current one

clients/iscmove/isc.go

PR #261, opened 2024-09-25
fix vmtests (wip)

dessaya:


howjmay (approved):

packages/coin/coin.go
```diff
@@ -1,46 +1,33 @@
...
 
 func (v Value) Bytes() []byte {
-	return rwutil.WriteToBytes(v)
+	return bcs.MustMarshal(&v)
 }
 
 func ValueFromBytes(b []byte) (Value, error) {
 	var r Value
-	_, err := rwutil.ReadFromBytes(b, &r)
+	r, err := bcs.Unmarshal[Value](b)
```
nnikolash: FYI thi can be just `return bcs.Unmarshal[Value](b)`

packages/coin/coin.go
```diff
@@ -1,46 +1,33 @@
...
 type Value uint64
 
-func (v Value) Write(w io.Writer) error {
-	ww := rwutil.NewWriter(w)
-	// serialized as bigint to save space for smaller values
-	ww.WriteBigUint(new(big.Int).SetUint64(uint64(v)))
-	return ww.Err
+func (v *Value) MarshalBCS(e *bcs.Encoder) error {
+	e.WriteCompactUint(uint64(*v))
+	return e.Err()
```
nnikolash, line 20: I was looking at this `coin.Value` type and did not migrate it because I was unsure how to correctly do it.

What you did is 100% correct in a way that it achieves same logic as it was before. But I was thinking about usage of this `Value` type in core contracts.

As far as I understand BCS only uses that "compact" integers (ULEB128) to store length of collections or IDs of enumation variants. I did not see a case, when they were using it just for a field of structure. So wouldn't that break our goal to make interface client-friendly?

So should we make it just a regular `uint64`? Also not sure about that too, because somebody at some point did this optimization for a reason.

That's why I stuck with it and not updated 😅 One of my ideas was that maybe it makes sense to add some serializer configuration, so that when serializing there will be different format for storage and interface...

nnikolash (approved):

packages/coin/coin.go
dessaya: The optimization was done thinking mainly of the space in the state database, where we store thousands if not millions of these values. The concern about client-friendliness does not apply there.

So maybe we should use a different type when exposing coin values to the client? cc @lmoe

PR #274, opened 2024-09-30
refactor: Limit the visibility of iscmove package

howjmay:


clients/iscmove/isc.go
```diff
@@ -93,6 +99,19 @@ func (a *Anchor) Equals(b *Anchor) bool {
 
 type AnchorWithRef = RefWithObject[Anchor]
 
+func AnchorWithRefEquals(a1 AnchorWithRef, a2 AnchorWithRef) bool {
+	if !a1.ObjectRef.Equals(&a2.ObjectRef) {
+		return false
+	}
+	if !reflect.DeepEqual(a1.Object.StateMetadata, a2.Object.StateMetadata) {
```
dessaya: `bytes.Equal`?

Also, maybe better to add `Equals()` to `Anchor`, and call it here.

Also, maybe better to add `Equals()` to `RefWithObject`, so that it can be called as `a1.Equals(a2)` instead of `AnchorWithRefEquals(a1, a2)`.

clients/iscmove/iscmoveclient/feed.go
```diff
@@ -228,3 +239,7 @@ func (f *ChainFeed) consumeAnchorUpdates(
 		}
 	}
 }
+
+func (f *ChainFeed) GetIscPackageID() sui.PackageID {
```
dessaya:
```suggestion
func (f *ChainFeed) GetISCPackageID() sui.PackageID {
```

packages/isc/sandbox_interface.go
```diff
@@ -301,6 +301,46 @@ type StateAnchor struct {
...
+	return &s.Ref.ObjectRef
+}
+
+func (s StateAnchor) GetObjectID() *sui.ObjectID {
+	return s.Ref.ObjectID
+}
+
+func (s StateAnchor) GetAssets() *Assets {
+	// FIXME
+	return nil
```
dessaya: Use `panic()` instead of `return nil` so that it's easier to find the error when it blows up in our face.

Also, as discussed IMO this method should not exist because we can't store all assets of the anchor in memory.

packages/nodeconn/chain.go
```diff
@@ -114,13 +115,19 @@ func (ncc *ncChain) postTxLoop(ctx context.Context) {
...
 func (ncc *ncChain) syncChainState(ctx context.Context) error {
 	ncc.LogInfof("Synchronizing chain state for %s...", ncc.chainID)
-	anchorRef, reqs, err := ncc.feed.FetchCurrentState(ctx)
+	moveAnchor, reqs, err := ncc.feed.FetchCurrentState(ctx)
 	if err != nil {
 		return err
 	}
-	ncc.anchorHandler(anchorRef)
+	// FIXME the owner should not be empty
+	anchor := isc.NewStateAnchor(moveAnchor, nil, ncc.feed.GetIscPackageID())
```
dessaya: Same here, please use panic() to make it easier to find the cause of the error.

Errors should never pass silently.

clients/iscmove/isc.go
```diff
@@ -83,6 +85,10 @@ type Anchor struct {
 	StateIndex    uint32
 }
 
+func (a *Anchor) GetStateMetadata() []byte {
```
lmoe: StateMetadata is a byte array, but actually an encoded struct and I think it already exists as StateMetadata in `isc`. Wondering if we should not just return the proper struct directly instead of just the byte array. @dessaya what do you think?

dessaya: Problem is that `isc` imports `iscmove`. This is a low-level struct.

I would remove this GetStateMetadata() method. Why is it needed?

howjmay: I added it because `GetStateIndex()` has existed. Therefore I thought it is better to make the pattern in consistent. Should I remove `GetStateIndex()` too?

dessaya: Given that `StateIndex` is already public I'd say yes, no need for the getter.

lmoe (approved):
lgtm i guess :D

howjmay:
love you so much

PR #285, opened 2024-10-03
refactor: Organize isc/move objects

howjmay:


clients/iscmove/isc.go
```diff
@@ -110,6 +99,43 @@ func (a1 Anchor) Equals(a2 *Anchor) bool {
...
 
+// test only
+type RandomAnchorOption struct {
+	ID            *sui.ObjectID
+	Assets        *AssetsBag
+	StateMetadata *[]byte
+	StateIndex    *uint32
+}
+
+func RandomAnchor(opt ...RandomAnchorOption) Anchor {
```
dessaya, line 110: IMO "test only" stuff should be in a separate package, like `iscmovetest`.

dessaya (approved):

clients/iscmove/isc.go
howjmay: fixed

PR #286, opened 2024-10-04
Fix/compilation tests

vitaliy-io:
did not touch foundries/native token creation


lmoe (approved):
Great! The concept of NFT as we have had it before does not exist anymore, which you correctly refactored with ObjectIDs. You did not change GetL2Nft to GetL2Object which I would keep as is for now. We can talk about that later. :)

nnikolash (commented):
I think you can replace `utxodb.FundsFromFaucetAmount` with `suiclient.FundsFromFaucetAmount` in `accounts_test.go`. I was not sure about it, but I've done it in my PR and Yanghao approved it 😅

PR #288, opened 2024-10-07
more fix

howjmay:


clients/iscmove/test_iscmove.go
```diff
@@ -0,0 +1,52 @@
+package iscmove
+
+// Everything in this file should be test only
```
dessaya: This is the same package `iscmove`.

What I proposed was moving this test code to a different package like `iscmovetest`, so that these functions are not available when you import `iscmove`.

howjmay: `iscmovetest` is created

sui-go/sui/test_sui.go
```diff
@@ -0,0 +1,23 @@
+package sui
+
+// TEST ONLY functions
+
+func RandomTransactionData() *TransactionData {
```
dessaya: Same here? :)

Test code should be separated from production code.

howjmay: moved!

dessaya (approved):

PR #300, opened 2024-10-09
Solo compiles and runs (badly)

dessaya:


packages/chainutil/evmestimategas.go
```diff
@@ -9,23 +9,29 @@ import (
...
 // for estimating gas.
 func EVMEstimateGas(
-	ch chain.ChainCore,
 	anchor *isc.StateAnchor,
+	store indexedstore.IndexedStore,
+	processors *processors.Cache,
+	log *logger.Logger,
 	call ethereum.CallMsg,
 ) (uint64, error) { //nolint:gocyclo,funlen
+	chainID := isc.ChainIDFromObjectID(*anchor.GetObjectID())
```
howjmay: I am thinking that whether adding a `GetChainID()` method for `isc.StateAnchor`, would make this easier and enhance the readability

packages/solo/solo.go
```diff
@@ -294,6 +292,56 @@ func (env *Solo) ISCPackageID() sui.PackageID {
 	return env.l1Config.ISCPackageID
 }
 
+func (env *Solo) pickCoinsForInitChain(
```
howjmay, line 295: Maybe we can use this to replace the `GetCoinObjsForTargetAmount ()` directly

howjmay (approved):

packages/solo/solo.go
dessaya: Don't know, this one is different because it creates a coin with _exactly_ the given amount. It's very specific to the start_new_chain function. Maybe it can be moved to the L2 client.

PR #302, opened 2024-10-10
fix suijsonrpc.Balance compilation

howjmay:


clients/iscmove/iscmoveclient/client_assets_bag.go
```diff
@@ -298,7 +298,7 @@ func (c *Client) GetAssetsBagWithBalances(
 		cointype := suijsonrpc.CoinType("0x" + data.Name.Value.(string))
 		bag.Balances[cointype] = &suijsonrpc.Balance{
 			CoinType:     cointype,
-			TotalBalance: moveBalance.Value.Uint64(),
+			TotalBalance: *suijsonrpc.NewBigInt(moveBalance.Value.Uint64()),
```
dessaya: why not make the field type `*BigInt` instead of `BigInt` to avoid having to dereference?

dessaya (approved):

PR #306, opened 2024-10-10
fix chainclient

howjmay:


clients/chainclient/chainclient.go
```diff
@@ -8,31 +8,39 @@ import (
...
 	"github.com/iotaledger/wasp/clients/apiclient"
 	"github.com/iotaledger/wasp/clients/apiextensions"
+	"github.com/iotaledger/wasp/clients/iscmove"
 	"github.com/iotaledger/wasp/packages/coin"
 	"github.com/iotaledger/wasp/packages/cryptolib"
 	"github.com/iotaledger/wasp/packages/isc"
 	"github.com/iotaledger/wasp/packages/kv/codec"
 	"github.com/iotaledger/wasp/packages/kv/dict"
 	"github.com/iotaledger/wasp/packages/vm/core/accounts"
+	"github.com/iotaledger/wasp/sui-go/sui"
```
lmoe: Please use iota-go now

lmoe (changes requested):

clients/iota-go/iotago/messages.go
```diff
@@ -73,6 +73,27 @@ func (p ProgrammableTransaction) PrintMoveCommand(cmdIdx int) {
 	}
 }
 
+func (p ProgrammableTransaction) IsInputObjectExist(targetID *ObjectID) bool {
```
dessaya: The name seems confusing. What does `Exist` mean here? Can it just be called `IsInputObject`?

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -96,6 +97,120 @@ func (c *Client) CreateAndSendRequest(
...
 
+func (c *Client) CreateAndSendRequestWithAssets(
+	ctx context.Context,
+	cryptolibSigner cryptolib.Signer,
+	packageID iotago.PackageID,
+	anchorAddress *iotago.ObjectID,
+	assets *isc.Assets,
+	iscContractHname uint32,
+	iscFunctionHname uint32,
+	args [][]byte,
```
dessaya: If `isc` is a lower level package then maybe these 3 should be `msg isc.Message`.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -96,6 +97,120 @@ func (c *Client) CreateAndSendRequest(
...
+	allowanceArray []iscmove.CoinAllowance,
```
dessaya: And this can just be `*isc.Assets`.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
 	"github.com/iotaledger/wasp/clients/iscmove"
 	"github.com/iotaledger/wasp/packages/cryptolib"
+	"github.com/iotaledger/wasp/packages/isc"
```
dessaya: Doesn't this create an import loop? I thought `iscmoveclient` was lower level than `isc`.

packages/testutil/l1starter/l1starter.go
```diff
@@ -110,7 +110,7 @@ func (in *IotaNode) execCmd() {
 		"--force-regenesis",
 		// TODO: remove when this is fixed: https://github.com/iotaledger/iota/issues/3212
 		// Run postgres with: docker run  -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432  postgres:latest
-		"--with-indexer", "--pg-user=postgres", "--pg-password=postgres", "--pg-db-name=postgres",
+		// "--with-indexer", "--pg-user=postgres", "--pg-password=postgres", "--pg-db-name=postgres",
```
dessaya: pls remove this

packages/vm/core/testcore/sbtests/sbtestsc/interface.go
```diff
@@ -16,7 +16,6 @@ var Processor = Contract.Processor(initialize,
 	FuncEventLogDeploy.WithHandler(testEventLogDeploy),
 	FuncSandboxCall.WithHandler(testSandboxCall),
 
-	FuncTestCustomError.WithHandler(testCustomError),
```
dessaya: why is this removed?

clients/iota-go/iotago/messages.go
howjmay: The `Inputs` field of `ProgrammableTransaction` is an array. So I want to express whether the object has existed in the array. For me `IsInputObject ` is a little bit confusing too, because is a ObjectID which for sure not a InputObject. Maybe I just need some time to get used to it

clients/iscmove/iscmoveclient/client_request.go
howjmay: `iscmoveclient` is not imported by `iscmove`, so the loop doesnt exist

clients/l2client.go
```diff
@@ -86,6 +87,20 @@ type L2Client interface {
 		ctx context.Context,
 		assetsBagID *iotago.ObjectID,
 	) (*iscmove.AssetsBagWithBalances, error)
+	CreateAndSendRequestWithAssets(
+		ctx context.Context,
+		cryptolibSigner cryptolib.Signer,
+		packageID iotago.PackageID,
+		anchorAddress *iotago.ObjectID,
+		assets *isc.Assets,
+		msg isc.Message,
```
dessaya: Ok but now there is an inconsistency. `CreateAndSendRequestWithAssets` receives parameters from `isc` package while e.g. `CreateAndSendRequest` does not (still receives `iscContractName uint32` instead of `isc.Message`).

I think this happens because we were not supposed to import `isc` from here.

howjmay: Should I revert `CreateAndSendRequestWithAssets` to the way `CreateAndSendRequest` is?

dessaya: IMO either don't use `isc`, or use it in all functions where it makes sense. Otherwise the inconsistency makes the package harder to use.

howjmay: I add similar types in `iscmove/isc.go`, so the interface may look better now

clients/iscmove/isc.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
+	"github.com/iotaledger/wasp/packages/coin"
```
dessaya: Same issue here: introducing the import `coin` significantly changes the dependency semantics. It is not supposed to be imported from here.

Either that, or maybe it makes sense to make `iscmove` higher-level than both `isc` and `coin`. But in that case we should refactor all types and methods to be defined in terms of `isc` stuff.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -96,6 +96,118 @@ func (c *Client) CreateAndSendRequest(
 	return txnResponse, nil
 }
 
+func (c *Client) CreateAndSendRequestWithAssets(
+	ctx context.Context,
+	cryptolibSigner cryptolib.Signer,
+	packageID iotago.PackageID,
+	anchorAddress *iotago.ObjectID,
+	assets *iscmove.Assets,
+	msg iscmove.Message,
```
dessaya: This introduces an inconsistency with `CreateAndSendRequest`, which receives `iscContractName uint32` etc.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -96,6 +96,118 @@ func (c *Client) CreateAndSendRequest(
...
+	getAllCoinsRes, err := c.GetAllCoins(ctx, iotaclient.GetAllCoinsRequest{Owner: cryptolibSigner.Address().AsIotaAddress()})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get anchor ref: %w", err)
+	}
+	coins := []*iotajsonrpc.Coin{}
+	// assume we can find it in the first page
+	for cointype, bal := range assets.Coins {
+		for _, coin := range getAllCoinsRes.Data {
+			if coin.CoinType == cointype.String() && coin.Balance.Uint64() > bal.Uint64() {
+				coin.Balance = iotajsonrpc.NewBigInt(bal.Uint64())
```
dessaya: Why change the coin's balance here?

packages/isc/assets.go
```diff
@@ -383,3 +383,16 @@ func (a *Assets) SetBaseTokens(amount coin.Value) *Assets {
 func (a *Assets) BaseTokens() coin.Value {
 	return a.Coins.Get(coin.BaseTokenType)
 }
+
+func (a *Assets) ToCoinAllowances() []iscmove.CoinAllowance {
```
dessaya: If this is no longer necessary then I'd remove it.

clients/iscmove/isc.go
howjmay: Isnt it ok to treat `coin` package as lower level or the same level as `iscmove`? The dependency of coin package is clear.

clients/iscmove/iscmoveclient/client_request.go
howjmay: I can change the interface of `CreateAndSendRequest `too.

clients/iscmove/iscmoveclient/client_request.go
howjmay: We will use the `coins` array as the parameter of `PTBAssetsBagPlaceCoinWithAmount()`. The balance of the coin is used as a separate parameter in that function to specify the amount to put into the AssetsBag.
And the ObjectRef in the coin object is used to make the coin object as a CallArg in the ptb. They are independent here

dessaya: IMO this is a bit hacky... It's confusing because it's changing the `Balance` of a `Coin` that has an `ObjectID`, which seems prone to subtle bugs. I think it's better if we can solve this without mutating the coins.

clients/iscmove/isc.go
dessaya: No, I think `coin` is on the same level as `isc`. It is separated only because of some import loop.

So IMO we have to choose:

1) Do not import `isc` or `coin` from here
2) Import them, but also refactor all types and methods to be defined in terms of `isc` stuff.

I'm fine either way, but not with something in the middle which is inconsistent.

Maybe @lmoe has a different opinion

lmoe: No I agree, it's important to find a decision between two and I would lean towards 1. As Diego said, coin was moved out of the isc package due to some import loop. 

iscmove/isc.go as you can see does not use any isc functions and types except in tests. -> For example we don't use isc.Hname but uint32s to have the least amount of coupling between isc internals.

I know it seems nicer and easier to just return a CoinBalance with typed values from that package but I think that will bite us later.

If I were you, I would create a type alias CoinBalances as map[string]uint64 and deal with the handling on the callers side. Maybe add a "NewAssetsFromCoinBalances" if we don't have it already. 

/cc @dessaya

clients/iscmove/iscmoveclient/client_request.go
howjmay: I created a new coin and append to the array `placedCoins`. I think this may be more clear

clients/iscmove/isc.go
howjmay: I have decoupled `coin` from `iscmove` package

clients/iscmove/isc.go
```diff
@@ -115,11 +115,44 @@ type Message struct {
...
+}
+
 type CoinAllowance struct {
 	CoinType iotajsonrpc.CoinType
 	Balance  uint64
 }
 
+type CoinBalances map[string]uint64
+
+var baseTokenType = "0x2::iota::IOTA"
```
dessaya: No need to redefine, it's in `iotajsonrpc.IotaCoinType`

clients/iota-go/iotago/messages.go
```diff
@@ -73,6 +73,15 @@ func (p ProgrammableTransaction) PrintMoveCommand(cmdIdx int) {
 	}
 }
 
+func (p ProgrammableTransaction) DoesObjectIdExistInInputObjects(targetID *ObjectID) bool {
```
dessaya: Nitpicking, but in Go the convention is `ID` instead of `Id`:

```suggestion
func (p ProgrammableTransaction) DoesObjectIDExistInInputObjects(targetID *ObjectID) bool {
```

Also, a bit shorter: `IsObjectIDInInputObjects`

Or even: `IsInInputObjects`

clients/iscmove/iscmoveclient/client_request.go
dessaya: The coin is still being mutated. I think you forgot to remove the line.

clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -96,6 +92,120 @@ func (c *Client) CreateAndSendRequest(
...
+	// assume we can find it in the first page
+	for cointype, bal := range assets.Coins {
+		for _, coin := range getAllCoinsRes.Data {
+			if coin.CoinType == cointype && coin.Balance.Uint64() > bal.Uint64() {
+				coin.Balance = iotajsonrpc.NewBigInt(bal.Uint64())
+				placedCoins = append(placedCoins, &iotajsonrpc.Coin{
+					CoinType:     cointype,
+					CoinObjectID: coin.CoinObjectID,
+					Balance:      iotajsonrpc.NewBigInt(bal.Uint64()),
+				})
```
dessaya: It's still strange that you are creating an instance of `iotajsonrpc.Coin`, which represents a L1 coin with an ObjectID, and assigning the ObjectID of an existing coin and a different balance. It doesn't really solve the problem that using `iotajsonrpc.Coin` for this purpose is incorrect.

IMO `placedCoins` should be a different type, for example:

```
placedCoins := []lo.Tuple[*iotajsonrpc.Coin, uint64]
```

i.e. a tuple where the first value is the original **unchanged** coin and the second the desired balance for splitting.

howjmay: fixed!

clients/iota-go/iotago/messages.go
howjmay: changed!

dessaya (approved):
:tada:

lmoe (approved):

lmoe:
Awesome :)

PR #315, opened 2024-10-14
Compilation fixes for the chain node.

kape1395:


kape1395:
I'm canceling it in favor of @lmoe PR.

PR #316, opened 2024-10-14
Iotago refactoring

lmoe:


clients/iota-go/contracts/isc/Move.toml
```diff
@@ -4,8 +4,6 @@ version = "0.0.1"
 edition = "2024.beta"
 
 [dependencies]
-MoveStdlib = { git = "git@github.com:iotaledger/iota.git", subdir = "crates/iota-framework/packages/move-stdlib", rev = "develop" }
```
howjmay, line 7: why do we remove this

lmoe: Doesn't seem to be required. Seems to be implied when importing the Iota dependency. I could build, deploy and use the contract without.

howjmay (approved):

clients/iota-go/iotasigner/signedtx.go
```diff
@@ -17,8 +17,13 @@ func NewSignedTransaction(unsignedTx *iotago.TransactionData, signature *Signatu
...
 
-func (st *SignedTransaction) Hash() hashing.HashValue {
-	panic("SignedTransaction.Hash not implemented") // TODO: Implement it.
+func (st *SignedTransaction) Hash() (hashing.HashValue, error) {
+	digest, err := st.Data.Digest()
+	if err != nil {
+		return hashing.NilHash, err
+	}
+
+	return hashing.HashValue(digest.Data()), nil
```
nnikolash, line 26: Did you verify that Digest will always return 32 bytes? If it is more - value will be truncated. If it is less - this line will panic.

Here is test: https://replit.com/@JleDeHeC/RectangularGrayWearables?v=1

lmoe: Damn this is actually a good catch, thanks for commenting on that again.

Seems like Digest is 44 bytes instead of 32. At least the ones I've seen. 

/edit

Digest is actually 32 bytes, but encoded with Base58 which is then a length of 44 or something. So we actually need to decode the Digest. Which means my function is still wrong, but at least we can keep using hashValue. :D

PR #322, opened 2024-10-16
fix mempool_test

howjmay:


packages/testutil/testchain/test_chain_ledger.go
```diff
@@ -45,132 +58,189 @@ func (tcl *TestChainLedger) ChainID() isc.ChainID {
...
+	// 	if output.Type() == iotago.OutputAlias {
+	// 		ao := output.(*iotago.AliasOutput)
+	// 		ao.StateIndex = anchor.GetStateIndex() + 1 // Fake next state index, just for tests.
+	// 		return isc.NewAliasOutputWithID(ao, outputID), tx
+	// 	}
+	// }
+	// panic("alias output not found")
+}
+
+func BuildAndDeployISCContracts(t *testing.T, client clients.L1Client, signer cryptolib.Signer) iotago.PackageID {
```
dessaya, line 219: This function seems copy-pasted? Searching for `contracts.ISC()` it seems like it is repeated like 4 times. It would be nice to avoid the duplication.

Also the name `BuildAndDeployISCContracts` is misleading because it does not actually build the contracts.

dessaya (approved):

packages/testutil/testchain/test_chain_ledger.go
howjmay: I think I will open another PR to merger all of them

PR #323, opened 2024-10-16
fix: Multiple compliction fixes related to transition from dict.Dict to isc.CallArguments

nnikolash:
Multiple compliction fixes related to transition from dict.Dict to isc.CallArguments


dessaya (approved):

clients/apiextensions/func.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"math"
+	"time"
+
+	iotago "github.com/iotaledger/iota.go/v3"
+	"github.com/iotaledger/wasp/clients/apiclient"
+	"github.com/iotaledger/wasp/packages/isc"
+)
+
+func CallArgsToAPIArgs(args isc.CallArguments) [][]int32 {
```
lmoe: Don't quite understand why you do this [][]int32 result here. I think []string (as hex encoded values) would work better if we send that to the API

lmoe (approved):

clients/apiextensions/func.go
nnikolash: `[][]int32` is generated, because `JSONDict` was replaced with `isc.CallArguments`:
```
	resExample := isc.NewCallArguments([]byte{1, 0, 7, 4})

	publicAPI.POST("chains/:chainID/callview", c.executeCallView).
		AddParamPath("", params.ParamChainID, params.DescriptionChainID).
		AddParamBody(mocker.Get(models.ContractCallViewRequest{}), "", "Parameters", true).
		AddResponse(http.StatusOK, "Result", resExample, nil).
		SetSummary("Call a view function on a contract by Hname").
		SetDescription("Execute a view call. Either use HName or Name properties. If both are supplied, HName are used.").
		SetOperationId("callView")
```

Although I did not introduced that change, I would agree that I did not think about it enough.
`JSONDict` has exactly yhe logic you said - hex-encoding of entries:
```
	for i, k := range d.KeysSorted() {
		j.Items[i].Key = iotago.EncodeHex([]byte(k))
		j.Items[i].Value = iotago.EncodeHex(d[k])
	}
```

I'll change my code to use that logic.

nnikolash: @lmoe I've updated review to use `[]string` instead of `[][]int32`.
Also, I see something was changed recently regarding `coin.Type`. Generated code also changed accordingly, but it does not seem for me as a good syntax - coin Type was string, now structure with field `S`. I think we would need to fix that

lmoe: Hmm. Yeah that won't work really well. For the WebAPI I would still expect a string==coin.Value so we would require a mapping type. I ask Diego what his intention was there.

lmoe: Ah I think I understand.

Yeah we should expect a string in the webapi requests, probably with a nicer alias like CoinValueAsString or something, and then map them into coin.Value where needed.

PR #332, opened 2024-10-21
Fixes for contract calls and more

vitaliy-io:


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -146,200 +142,6 @@ func TestWaspCLIDeposit(t *testing.T) {
 		checkBalance(t, w.MustRun("chain", "balance", eth.String(), "--node=0", "--address-index=1"), 1000000) // fee will be taken from the sender on-chain balance
 		checkBalance(t, w.MustRun("chain", "balance", "--node=0", "--address-index=1"), 1000000-2*int(minFee))
 	})
-
-	t.Run("mint and deposit native tokens to an ethereum account", func(t *testing.T) {
```
dessaya, line 150: Why is this test removed instead of commented out?

vitaliy-io: This should be commented, you are right. Fixed

dessaya (approved):

PR #342, opened 2024-10-30
Fix compilation errors, fix webapi calls, fix wasp-cli calls, decoding issues

lmoe:
* Fixes the iscmove contract to alphanet 0.4.0
* Updates the Multiclient to support multiple response tx types
* Fixes some webapi calls that are still using ChainID where they can't
* Fix some tests breaking go mod (still including kvdecoder that was removed)
* Fix some wasp-cli commands, some are set to ignore (minting coins, deploying contract) some are updated to isc.CallArguments


tools/wasp-cli/chain/accounts.go
```diff
@@ -46,7 +47,7 @@ func initBalanceCmd() *cobra.Command {
 
 			rows[0] = []string{"base", balance.BaseTokens}
 			for k, v := range balance.NativeTokens {
-				rows[k+1] = []string{v.Id, v.Amount}
+				rows[k+1] = []string{v.CoinType.GetS(), ""} // Place balance here, (requires fix on the webapi to return a bigint as string)
```
dessaya: IMO it is better to `panic` until it is fixed.

clients/apiextensions/func.go
```diff
@@ -2,6 +2,7 @@ package apiextensions
 
 import (
 	"fmt"
+	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
 	"time"
```
dessaya: Your IDE configuration places the imports in a different order?

dessaya (approved):

clients/apiextensions/func.go
lmoe: Sigh, yeah I had to reset my IDE as it was fully broken and forgot to configure import rules. Sry!

PR #345, opened 2024-11-01
Evm contract call from vm test

cwarnerdev:
- adds a test to call magic contract method from within vm.
- fixes a nil pointer edge case
- disabled some evmtest code that is not working. will investigate for further improvement/fixes

### DIFF
--- a/packages/chainutil/vm_test.go
+++ b/packages/chainutil/vm_test.go
@@ -1,4 +1,4 @@
-package evmtest
+package chainutil
 
 import (
 	"bytes"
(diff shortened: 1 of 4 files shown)


packages/vm/vmimpl/call.go
```diff
@@ -43,10 +43,12 @@ const traceStack = false
...
-		contract:           contract,
-		params:             args,
-		allowanceAvailable: allowance.Clone(), // we have to clone it because it will be mutated by TransferAllowedFunds
+		caller:   caller,
+		contract: contract,
+		params:   args,
+	}
+	if allowance != nil {
+		ctx.allowanceAvailable = allowance.Clone() // we have to clone it because it will be mutated by TransferAllowedFunds
 	}
```
dessaya: We changed `Assets` so that it should never be nil. So instead of checking here it is better to find the root cause of why it is nil and fix it there.

cwarnerdev:
No more nil check for the allowance :tada:

dessaya (approved):

PR #369, opened 2024-11-20
feat: take coin from asset bag to top up gas coin

howjmay:


packages/isc/sandbox_interface.go
```diff
@@ -363,20 +365,44 @@ type StateAnchor struct {
...
+		GasCoin:    gasCoin,
 	}
 }
 
+// Every time changing the L1 state of the Anchor object, the nodes should call
+// this func to synchronize to the latest state
+// It includes the following 'txbuilder' operations
+// * BuildTransactionEssence (update the anchor commitment)
+// * RotationTransaction
+func (a *StateAnchor) UpdateStateAnchor(ctx context.Context, c clients.L1Client) error {
```
dessaya: I don't think mutating the StateAnchor is a good idea. IMO it is better to just fetch the latest version of the anchor and create a new StateAnchor instance.

clients/iscmove/iscmoveclient/client_assets_bag.go
```diff
@@ -228,6 +228,67 @@ func (c *Client) AssetsDestroyEmpty(
...
+		txnBytes,
+		&iotajsonrpc.IotaTransactionBlockResponseOptions{ShowEffects: true, ShowObjectChanges: true},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("can't execute the transaction: %w", err)
+	}
+	if !txnResponse.Effects.Data.IsSuccess() {
+		return nil, fmt.Errorf("failed to execute the transaction: %s", txnResponse.Effects.Data.V1.Status.Error)
+	}
+	return txnResponse, nil
```
dessaya: With my latest PR merged this can be replaced with `c.SignAndExecutePTB()`

howjmay: yes yes. Thank a lot. That was a pain point I always want to solve. I even opened a issue for it

packages/isc/sandbox_interface.go
howjmay: I removed this func, and amended the comment for using in `NewStateAnchor()`

clients/iscmove/iscmoveclient/client_anchor.go
```diff
@@ -83,6 +83,8 @@ func (c *Client) ReceiveRequestsAndTransition(
 	anchorRef *iotago.ObjectRef,
 	reqs []iotago.ObjectRef,
 	stateMetadata []byte,
+	gasCoinRef *iotago.ObjectRef,
+	topUpAmount uint64,
 	gasPayments []*iotago.ObjectRef, // optional
```
dessaya: So the gasCoin is received twice? I mean `gasCoinRef` and `gasPayments`. Shouldn't it be the same coin?

howjmay: Yes. I think I didnt properly set up the test. Now only the gasPayment exist. 
but I change the type from a slice into a single ObjectRef. Not sure whether it make it hard to understand?

I think the current `TestReceiveRequestAndTransition()` can ease your consideration about the we cant mutate the GasCoin. You can see I am using GasCoin in `PTBAssetsBagTakeCoinBalanceMergeTo()`

packages/vm/vmtxbuilder/txbuilder_test.go
```diff
@@ -98,37 +103,49 @@ func TestTxBuilderSendAssetsAndRequest(t *testing.T) {
...
+	// txnBytes1, err := bcs.Marshal(&tx1)
+	// require.NoError(t, err)
+
+	// txnResponse1, err := client.SignAndExecuteTransaction(
+	// 	context.Background(),
+	// 	cryptolib.SignerToIotaSigner(chainSigner),
+	// 	txnBytes1,
+	// 	&iotajsonrpc.IotaTransactionBlockResponseOptions{ShowEffects: true, ShowObjectChanges: true},
+	// )
+	// require.NoError(t, err)
```
dessaya: this intentionally left commented out?

dessaya (approved):

packages/vm/vmtxbuilder/txbuilder_test.go
howjmay: Sorry it was my mistake. I have removed it, and I fixed the txbuilder_test too

PR #370, opened 2024-11-20
accounts_test.go done

dessaya:


clients/iscmove/iscmoveclient/client_anchor.go
```diff
@@ -7,25 +7,54 @@ import (
 	"github.com/iotaledger/wasp/clients/iota-go/iotaclient"
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
-	"github.com/iotaledger/wasp/packages/util/bcs"
 
 	"github.com/iotaledger/wasp/clients/iscmove"
 	"github.com/iotaledger/wasp/packages/cryptolib"
 )
 
+func (c *Client) FindCoinsForGasPayment(
```
howjmay, line 15: Do we need to check weather the chosen Coin here has been used in the PTB?
We already have some similar func like `FindCoinsForGasPayment `. But the collision of Coin is more annoying

dessaya: > Do we need to check weather the chosen Coin here has been used in the PTB?

Yes, otherwise the PTB fails.

> We already have some similar func like FindCoinsForGasPayment

Which one?

howjmay (approved):

clients/iscmove/iscmoveclient/client_anchor.go
howjmay: This one `iotajsonrpc.PickupCoins()`. But I dont mind deleting `iotajsonrpc.PickupCoins()` and replace it with this one

howjmay: Just both funcs dont solve the collision error. In some tests there are too many Coins are used. It may be hard to understand things there

dessaya: Ah ok, there is also `GetCoinObjsForTargetAmount` which calls `PickupCoins` internally.

But I don't think I can use either because I need to make sure the selected coins are not referenced in the PTB.

Maybe I can modify `PickupCoins` to receive an optional filter function...

> Just both funcs dont solve the collision error

What collision do you mean?

howjmay: You have solved it. I meant the input of PTB.

PR #372, opened 2024-11-21
chore: refactoring

lmoe:


packages/isc/sandbox_interface.go
```diff
@@ -402,6 +402,10 @@ func (s StateAnchor) Hash() hashing.HashValue {
 }
 
 func (s StateAnchor) Equals(input *StateAnchor) bool {
+	if input == nil {
+		return false
+	}
+
```
dessaya, line 408: This looks fishy. Is it considered ok for the pointer to be nil?

tools/wasp-cli/chain/deploy.go
```diff
@@ -5,6 +5,9 @@ package chain
 
 import (
 	"context"
+	"github.com/iotaledger/hive.go/kvstore/mapdb"
+	"github.com/iotaledger/wasp/packages/state/indexedstore"
+	"github.com/iotaledger/wasp/packages/transaction"
 	"os"
```
dessaya: Pls see if you can tell your IDE to place the imports in the right order :rofl:

dessaya (approved):

tools/wasp-cli/chain/deploy.go
lmoe: Jesus christ, not again

packages/isc/sandbox_interface.go
lmoe: It's getting used quite often and the value is directly passed into it. Sometimes the pointer can be nil it seems.

lmoe: Mostly in the consensus / state parts. Where we have callbacks receiving a pointer. In some cases it was nil there, yeah

tools/wasp-cli/chain/deploy.go
lmoe: fixed :D

clients/iscmove/iscmoveclient/client_anchor.go
```diff
@@ -37,7 +37,8 @@ func (c *Client) FindCoinsForGasPayment(
 
 func (c *Client) StartNewChain(
 	ctx context.Context,
-	signer cryptolib.Signer,
+	cryptolibSigner cryptolib.Signer,
+	chainOwnerAddress *cryptolib.Address,
```
howjmay, line 41: So the signer may not be the be the chain owner?

lmoe: Yes, if you call this function with the wasp-cli you call it with your own key, not the Committees key. Which requires you to transfer the ownership of the Anchor afterwords. 

https://github.com/iotaledger/isc-private/blob/develop/tools/wasp-cli/chain/deploy.go#L121
https://github.com/iotaledger/isc-private/blob/develop/tools/wasp-cli/chain/deploy.go#L134

PR #374, opened 2024-11-21
feat: Add DevInspectPTB for inspect mode

howjmay:


clients/iscmove/iscmoveclient/client_request.go
```diff
@@ -64,6 +64,60 @@ func (c *Client) SignAndExecutePTB(
...
+		return nil, fmt.Errorf("can't execute the transaction: %w", err)
+	}
+	if txnResponse.Error != "" {
+		return nil, fmt.Errorf("execute error: %s", txnResponse.Error)
+	}
+	if !txnResponse.Effects.Data.IsSuccess() {
+		return nil, fmt.Errorf("failed to execute the transaction: %s", txnResponse.Effects.Data.V1.Status.Error)
+	}
+	return txnResponse, nil
+}
```
dessaya: This looks almost exactly like `SignAndExecutePTB`. Is it possible to avoid the code duplication?

Seems like the same calling logic could be reused for `iota_devInspectTransactionBlock`, `iota_dryRunTransactionBlock` and `iota_executeTransactionBlock`.

howjmay: I think it is a little tricky 

The API interfaces of these three APIs are slightly different from each other.

I think only the following part is exactly the same. If we refactor them to a more general interface, then I am afraid of making them even more complex.

```
if len(gasPayments) == 0 {
		gasPayments, err = c.FindCoinsForGasPayment(
			ctx,
			signer.Address(),
			pt,
			gasPrice,
			gasBudget,
		)
		if err != nil {
			return nil, err
		}
	}

	tx := iotago.NewProgrammable(
		signer.Address(),
		pt,
		gasPayments,
		gasBudget,
		gasPrice,
	)
```

howjmay:
I add one more commit to synchronize the interface of SignAndExecuteTransaction as other APIs too

dessaya (approved):

PR #376, opened 2024-11-22
fix: Add a constant top up fee in VM

howjmay:


lmoe (approved):

packages/vm/vmimpl/gas.go
```diff
@@ -1,12 +1,16 @@
...
 import (
+	"github.com/iotaledger/wasp/clients/iota-go/iotaclient"
 	"github.com/iotaledger/wasp/packages/coin"
 	"github.com/iotaledger/wasp/packages/vm"
 	"github.com/iotaledger/wasp/packages/vm/gas"
 	"github.com/iotaledger/wasp/packages/vm/vmexceptions"
 )
 
+// TODO Add the comprehensive top up calculation logic, then we can remvoe this constant
+const TopUpFee = iotaclient.DefaultGasBudget * 5
```
dessaya, line 12: DefaultGasBudget is expressed in "gas units"? If that is the case this should be:

```
iotaclient.DefaultGasBudget * iotaclient.DefaultGasPrice * 5
```

howjmay: I think its unit is mist (in Sui basetoken) as the document mentions

https://docs.sui.io/concepts/tokenomics/gas-in-sui#gas-budgets

PR #377, opened 2024-11-25
fix: Fixes to webapi

nnikolash:


packages/webapi/common/vm.go
```diff
@@ -14,66 +15,77 @@ import (
 )
 
 func ParseReceipt(chain chainpkg.Chain, receipt *blocklog.RequestReceipt) (*isc.Receipt, error) {
-	panic("refactor me: ParseReceipt")
-	/*resolvedReceiptErr, err := chainutil.ResolveError(chain, receipt.Error)
+	state, err := chain.Store().StateByIndex(receipt.BlockIndex)
```
nnikolash: This function parses request receipt. On this line we retrieve state, corresponring to the request in receipt. The state is used to resolve error message.

Is it possible, that request is related to state which is too "old"? If yes, should we still support that? For that case, does it make sense to use `chain.LatestState(chainpkg.ActiveOrCommittedState)` instead of `StateByIndex()`?

packages/webapi/common/vm.go
```diff
@@ -14,66 +15,77 @@ import (
...
 
 	iscReceipt := receipt.ToISCReceipt(resolvedReceiptErr)
 
-	return iscReceipt, nil*/
+	return iscReceipt, nil
 }
 
 func CallView(ch chainpkg.Chain, msg isc.Message, blockIndexOrHash string) (isc.CallArguments, error) {
-	panic("refactor me: CallView")
+	anchor, err := ch.LatestAnchor(chainpkg.ActiveOrCommittedState)
```
nnikolash: On this line we retrieve latest anchor, even though the function has `blockIndexOrHash` argument.
This is because the anchor is used later only to get **chain ID**  (in `chainutil.CallView() -> viewcontext.New()`).

So is it ok to use same anchor? Or is it still cleaner to retrieve anchor object of proper version usig `GetPastAnchorFromObjectID()`?

Or if using `GetPastAnchorFromObjectID()` is an overkill, maybe it makes sense to change `chainutil.CallView()` and `viewcontext.New()` to accept **chain ID** instead of anchor?

lmoe: As we rely on the ID and not the ObjectRef, I think your solution is sensible here. The ID does not change. 

Your point regarding viewcontext.New is correct though, I agree, we could refactor it in this case. :)

packages/webapi/common/vm.go
lmoe: I guess you are right, technically. The error format that we parse does not really change, so we can always use the Active state I think.

nnikolash: Although it does not change in realife usage, should we still support such change?

packages/webapi/common/vm.go
nnikolash: Ok, then I'll `viewcontext.New` and `chainutil.CallView()` and will see how much changes it would require.

packages/webapi/common/vm.go
lmoe: I think you should try it with the Active state. Even though using StateByIndex "feels" more correct as you want to pull the state you are asking for in the receipt.

But we do prune after a while and then the state is gone. If its just related to the error decoding, the active state should be fine. Try that. :)

lmoe (approved):

PR #383, opened 2024-11-25
chore: generic refactoring, remove NetworkPrefix

lmoe:
* Replace iota.go EncodeHex => hexutil.Encode
* Fix reading chain state by Hex address
* Remove NetworkPrefix

### DIFF
--- a/packages/isc/chainid.go
+++ b/packages/isc/chainid.go
@@ -4,7 +4,6 @@
 package isc
 
 import (
-	iotago2 "github.com/iotaledger/iota.go/v3"
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 	"github.com/iotaledger/wasp/packages/cryptolib"
 	"github.com/iotaledger/wasp/packages/util/bcs"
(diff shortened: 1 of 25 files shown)


packages/isc/chainid.go
```diff
@@ -83,7 +78,7 @@ func (id ChainID) Equals(other ChainID) bool {
 }
 
 func (id ChainID) Key() ChainIDKey {
-	return ChainIDKey(id.AsAliasID().String())
+	return ChainIDKey(id.AsObjectID().String())
 }
```
dessaya, line 82: This ChainIDKey should not be necessary anymore, because ChainID is now an array and can be used as-is as a map key.

dessaya (approved):

packages/isc/chainid.go
lmoe: Yeah for now I just refactored this little piece as ChainKey was used quite often somewhere else and I already spent enough time yesterday. :D 

I'll update this today.

PR #386, opened 2024-11-25
refactor: Unify the coin type usage with coin.Type

### LINKED ISSUE #353: coin.Type may be inconsistent in the same coin
`iotajsonrpc.CoinType` and `coin.Type` are alias of string or derivation of string. However, the coin type in Sui has two format `long` and `short`. We need to ensure that all the coin type have been parsed before the following process

howjmay:
close [#353](https://github.com/iotaledger/isc-private/issues/353)

### DIFF
--- a/clients/iscmove/isc.go
+++ b/clients/iscmove/isc.go
@@ -63,7 +63,7 @@ func (ab *AssetsBag) Equals(other *AssetsBag) bool {
 		ab.Size == other.Size
 }
 
-type AssetsBagBalances map[iotajsonrpc.CoinType]*iotajsonrpc.Balance
+type AssetsBagBalances map[iotajsonrpc.CoinType]iotajsonrpc.CoinValue
 
 type AssetsBagWithBalances struct {
 	AssetsBag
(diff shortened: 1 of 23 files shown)


clients/iscmove/isc.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bytes"
 
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
-	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
+	"github.com/iotaledger/wasp/packages/coin"
```
lmoe: I think we had this discussion already a while ago about importing isc packages like coin from the clients. Lets talk about it tomorrow.

clients/iscmove/iscmoveclient/client.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/iotaledger/wasp/clients/iota-go/iotaclient"
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 	"github.com/iotaledger/wasp/clients/iota-go/iotajsonrpc"
+	"github.com/iotaledger/wasp/packages/coin"
```
dessaya: This adds a dependency to `coin` from `iscmoveclient`, which we were avoiding.

clients/iscmove/iscmoveclient/client.go
```diff
@@ -161,3 +162,7 @@ func (c *Client) DevInspectPTB(
 	}
 	return txnResponse, nil
 }
+
+func IscCointypeToIotagoRpcCointype(c coin.Type) iotajsonrpc.CoinType {
```
dessaya: So maybe this function should be in `coin` package.

packages/coin/coin.go
```diff
@@ -131,6 +131,10 @@ func TypeFromBytes(b []byte) (Type, error) {
 	return r, err
 }
 
+func (t Type) IotagoRpcCointype() iotajsonrpc.CoinType {
```
dessaya:
```suggestion
func (t Type) IotagoRPCCoinType() iotajsonrpc.CoinType {
```

Also I would remove `CoinType` from the name to avoid stuttering.

```suggestion
func (t Type) ToIotaJSONRPC() iotajsonrpc.CoinType {
```

dessaya (approved):

PR #397, opened 2024-11-29
fix misuse of StateAnchor

howjmay:


packages/chain/cmt_log/var_localview_rapid_test.go
```diff
@@ -51,113 +45,113 @@ func newVarLocalViewSM(t *rapid.T) *varLocalViewSM {
 	return sm
 }
 
-// E.g. external rotation of a TX by other chain.
```
howjmay, line 54: All these are not used, so I comment them

packages/isc/isctest/random.go
```diff
@@ -1,36 +1,82 @@
...
-	anchor := iscmovetest.RandomAnchor()
-	anchorRef := iscmove.RefWithObject[iscmove.Anchor]{
-		ObjectRef: *iotatest.RandomObjectRef(),
-		Object:    &anchor,
-		Owner:     cryptolib.NewRandomAddress().AsIotaAddress(),
+// simualte how StateAnchor is updated (state transition)
+// assume the AssetsBag keep unchanged
+func UpdateStateAnchor(anchor *isc.StateAnchor, stateMetadata ...[]byte) *isc.StateAnchor {
+	anchor.Anchor().ObjectRef.Version += 1
+	anchor.Anchor().Object.StateIndex += 1
```
dessaya: I don't think we should mutate objects that are not meant to be mutated.

howjmay: I dont understand. I think it is ok to do in this test only function.
 If we want to transit the Anchor's state, then bother ObjectRef.Version and StateIndex have to increment.

dessaya: What I mean is that some objects are immutable and some are not. In languages like Rust you can forbid mutations in immutable objects, but that is not possible in Go, so we must enforce that by convention.

In that sense, StateAnchor is immutable as it can only be constructed from an Anchor retrieved from L1 (which is also immutable). So by convention we should not mutate it, and if we need to "change" it we must return a new instance with the applied changes. IMO we should adhere to this even in tests, for the sake of consistency and maintainability.

lmoe: Fully agree on that. I made the same mistake after the stateanchor refactor and now return a new copy rather.

howjmay: Got it. I misunderstood about your idea at the beginning. sorry. 
I have used Clone() to copy the object and update the generated object

clients/iota-go/iotago/address.go
```diff
@@ -95,3 +95,9 @@ func (a *Address) UnmarshalJSON(data []byte) error {
 func (a Address) MarshalBCS() ([]byte, error) {
 	return a[:], nil
 }
+
+func (a Address) Clone() Address {
+	var ret [AddressLen]uint8
+	ret = a
+	return ret
+}
```
dessaya:
```suggestion
func (a Address) Clone() (ret Address) {
	ret = a
	return
}
```

Also, aren't we passing `Address` as pointer everywhere?

clients/iota-go/iotago/base_types.go
```diff
@@ -63,6 +63,21 @@ func (o ObjectRef) Hash() hashing.HashValue {
 	return res
 }
 
+func (o ObjectRef) Clone() *ObjectRef {
```
dessaya: Same here, shouldn't the receiver be pointer?

Also, "cloning" an immutable object should never be necessary. It defeats the purpose of making the object immutable.

packages/chain/cmt_log/cmt_log_rapid_test.go
```diff
@@ -74,26 +72,21 @@ func newCmtLogTestRapidSM(t *rapid.T) *cmtLogTestRapidSM {
...
-	stateAnchor := isc.NewStateAnchor(
-		&iscmove.AnchorWithRef{
-			Object:    &anchor,
-			ObjectRef: sm.anchorRef,
-			Owner:     sm.stateAddress.AsIotaAddress(),
-		}, *cryptolib.NewRandomAddress().AsIotaAddress())
-
+// simulate StateAnchor to state transition
+func (sm *cmtLogTestRapidSM) nextStateAnchorWithStateIndex(stateIndex uint32) *isc.StateAnchor {
+	sm.anchorRef.Version += 1
```
dessaya: This is mutating the AnchorRef.

Just create a new one with the updated values.

packages/isc/isctest/random.go
```diff
@@ -1,36 +1,83 @@
...
-	anchorRef := iscmove.RefWithObject[iscmove.Anchor]{
-		ObjectRef: *iotatest.RandomObjectRef(),
-		Object:    &anchor,
-		Owner:     cryptolib.NewRandomAddress().AsIotaAddress(),
+// simualte how StateAnchor is updated (state transition)
+// assume the AssetsBag keep unchanged
+func UpdateStateAnchor(anchor *isc.StateAnchor, stateMetadata ...[]byte) *isc.StateAnchor {
+	a := anchor.Clone()
+	a.Anchor().ObjectRef.Version += 1
+	a.Anchor().Object.StateIndex += 1
```
dessaya: Same here, just create a new anchor with the updated values. You can create a function for that if you want, but it should be test-only code, since the object is immutable and in production it will never be necessary to clone it or modify its values.

clients/iota-go/iotago/address.go
```diff
@@ -95,3 +95,8 @@ func (a *Address) UnmarshalJSON(data []byte) error {
 func (a Address) MarshalBCS() ([]byte, error) {
 	return a[:], nil
 }
+
+func (a Address) Clone() (ret Address) {
+	ret = a
+	return
+}
```
dessaya: Since Address is also immutable this function should not be necessary.

packages/chain/cmt_log/cmt_log_rapid_test.go
```diff
@@ -74,26 +72,26 @@ func newCmtLogTestRapidSM(t *rapid.T) *cmtLogTestRapidSM {
...
-			Object:    &anchor,
-			ObjectRef: sm.anchorRef,
-			Owner:     sm.stateAddress.AsIotaAddress(),
-		}, *cryptolib.NewRandomAddress().AsIotaAddress())
-
+// simulate StateAnchor to state transition
+func (sm *cmtLogTestRapidSM) nextStateAnchorWithStateIndex(stateIndex uint32) *isc.StateAnchor {
+	newObjID := sm.anchorRef.ObjectID.Clone()
+	newAnchor := iotago.ObjectRef{
+		ObjectID: &newObjID,
```
dessaya:
```suggestion
		ObjectID: &sm.anchorRef.ObjectID,
```

(no need to clone it since it's immutable).

dessaya (approved):

PR #405, opened 2024-12-02
remove some iota.go/v3 dependency

howjmay:


packages/parameters/l1parameters.go
```diff
@@ -58,18 +49,6 @@ var (
 	L1ForTesting = &L1Params{
 		// There are no limits on how big from a size perspective an essence can be, so it is just derived from 32KB - Message fields without payload = max size of the payload
 		MaxPayloadSize: MaxPayloadSize,
-		Protocol: &iotago.ProtocolParameters{
```
lmoe: What about the ProtocolParams here?

I've thought they can change every Epoch, wouldnt it make sense to refactor this place here?

howjmay: I thinking to refactor that in the following PR to make the incoming PR smaller. I can remove this change to make this PR contain less things

howjmay: removed

lmoe: Okay, both ways are fine. Just wanted to ask. :D

howjmay: I just removed the mock_ledger too. Please check. Sorry

dessaya (approved):

PR #407, opened 2024-12-03
fix: Fixed GetControlAddresses

nnikolash:


packages/webapi/corecontracts/blocklog.go
```diff
@@ -8,17 +8,15 @@ import (
...
-	// 	GoverningAddress: cryptolib.NewAddressFromIota(aliasOutput.GovernorAddress()),
-	// 	SinceBlockIndex:  aliasOutput.StateIndex,
-	// }, nil
+	anchor, err := ch.LatestAnchor(chain.ConfirmedState)
+	if err != nil {
+		return nil, err
+	}
+	return &isc.ControlAddresses{
+		StateAddress:     ch.GetCommitteeInfo().Address,
+		GoverningAddress: ch.GetCommitteeInfo().Address,
```
dessaya: If they are the same address, does it make sense to have two different fields?

dessaya (approved):

packages/webapi/corecontracts/blocklog.go
nnikolash: IDK. As far as I understand we had some of these decisions just for a temporary simplicity. So it has a high chance to changing in the not-so-distant future, right?
And to fix this, we would need to change web api, which overall requires a lot more review than just this. I suspect we will do that review eventually later on?

PR #413, opened 2024-12-05
fix: integration tests

lmoe:
# Description of change

This makes use of Docker and testcontainers to properly spin iota nodes up and down for integration tests.

It should solve quite a few issues but may create a few new ones.

* All tests should run on all OSes as the iota node is dockerized and does not need to run directly on the local machine
* All packages now bootstrap their own node (by starting a separate container and proxying the default 9000/9123 port to dynamic ones per package)

Unfortunately, this means you now need docker :D 
It probably is now slower than before but we can improve this still.

You can create a `.testconfig` file to change the test behavior / L1 endpoint:

```
{
    "IS_LOCAL": false,
    "API_URL": "http://localhost:9000/",
    "FAUCET_URL": "http://localhost:9123/gas"
}
```

`IS_LOCAL` = `true` will always start a local iota node via Docker and disregard the URLs.
Set it to false, to use an external node like Alphanet.


clients/iota-go/iotaclient/faucet.go
```diff
@@ -13,7 +13,11 @@ import (
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 )
 
-const FundsFromFaucetAmount = 9_999_999_999_999
+// We can set a certain amount of coin returned from the faucet. However,
+// you get this value 5 times with 5 coins. Therefore,
+// we need to assert account balances faucetamount * 5
+const SingleCoinFundsFromFaucetAmount = 200_000_000_0
```
dessaya:
```suggestion
const SingleCoinFundsFromFaucetAmount = 2_000_000_000
```

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
```diff
@@ -15,14 +17,26 @@ import (
...
 	seed[0] = seed[0] + byte(index)
 	kp := cryptolib.KeyPairFromSeed(cryptolib.Seed(seed))
-	err := iotaclient.RequestFundsFromFaucet(context.TODO(), kp.Address().AsIotaAddress(), iotaconn.LocalnetFaucetURL)
+	err := iotaclient.RequestFundsFromFaucet(context.TODO(), kp.Address().AsIotaAddress(), l1starter.Instance().FaucetURL())
 	require.NoError(t, err)
 	return kp
 }
 
-func NewLocalnetClient() *iscmoveclient.Client {
+func NewIscMoveWebSocket(ctx context.Context, log *logger.Logger) (*iscmoveclient.Client, error) {
```
dessaya:
```suggestion
func NewISCMoveWebSocket(ctx context.Context, log *logger.Logger) (*iscmoveclient.Client, error) {
```

Also, since this function is already in `iscmoveclient` package, I would name it something like `NewWebSocketClient` to avoid stuttering.

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
```diff
@@ -15,14 +17,26 @@ import (
...
+		ctx,
+		iotaconn.AlphanetWebsocketEndpointURL,
+		l1starter.Instance().FaucetURL(),
+		log,
+	)
+
+	return wsClient, err
+}
+
+func NewIscMoveClient() *iscmoveclient.Client {
```
dessaya: `NewClient` or `NewHTTPClient` or even just `New`

packages/solo/solo.go
```diff
@@ -6,7 +6,7 @@ package solo
 import (
 	"bytes"
 	"context"
-	"fmt"
+	"github.com/iotaledger/wasp/packages/testutil/l1starter"
```
dessaya: Again your IDE placing imports in weird places :rofl:

dessaya (approved):

packages/solo/solo.go
lmoe: This time my IDE is innocent. :D

PR #415, opened 2024-12-05
Add missing method for chain.Chain{} and fix offledger_test

howjmay:


packages/solo/chain.go
```diff
@@ -30,10 +38,144 @@ import (
...
+	log *logger.Logger
+	// global processor cache
+	proc *processors.Config
+	// related to asynchronous backlog processing
+	runVMMutex sync.Mutex
+
+	migrationScheme *migrations.MigrationScheme
+}
+
+var _ chain.Chain = &Chain{}
```
dessaya, line 70: Why is it necessary to implement `chain.Chain` here?

howjmay: It is required here at `packages/webapi/test/offledger_test.go`

```
rec, err := common.EstimateGas(chain, altReq)
```

howjmay: I push another commit to fix the offledger_test. It is a little bit more than the first commit. If that commit is confusing in this PR, I will put in another one

dessaya: > It is required here at `packages/webapi/test`

Then it would be nice if we can refactor `packages/webapi/test` to not depend on `chain.Chain`. Maybe depend on a smaller interface.

The reason is that it does not make sense for Solo to implement the whole interface, especially if most methods are panicking.

PR #417, opened 2024-12-05
fix: Getting anchor by state index

nnikolash:


packages/webapi/services/waspevmbackend.go
```diff
@@ -33,7 +34,19 @@ type WaspEVMBackend struct {
 }
 
 func (b *WaspEVMBackend) ISCAnchor(stateIndex uint32) (*isc.StateAnchor, error) {
-	panic("refactor me: ISCAnchor (get state by state index, return anchor)")
+	state, err := b.chain.LatestState(chain.ConfirmedState)
```
nnikolash, line 38: Is it right, that I've using latest state? I've done it, because I suspect db states are pruned, while blocklog state is kept (or blocklog just lives longer?).

packages/vm/vmimpl/vmcontext.go
```diff
@@ -142,6 +142,7 @@ func (vmctx *vmContext) saveBlockInfo(numRequests, numSuccess, numOffLedger uint
 		SchemaVersion:         blocklog.BlockInfoLatestSchemaVersion,
 		BlockIndex:            vmctx.stateDraft.BlockIndex(),
 		Timestamp:             vmctx.stateDraft.Timestamp(),
+		Anchor:                vmctx.task.Anchor,
```
nnikolash: I'm storing the entire state anchor, because it is returned from the function to be fixed, and I don't know if we can retrieve state anchor just by having anchor obj ref and package ID.
If we can and it makes sense - please hint me, I'll change it.

packages/vm/core/blocklog/impl.go
```diff
@@ -22,6 +22,8 @@ var Processor = Contract.Processor(nil,
 var ErrBlockNotFound = coreerrors.Register("Block not found").Create()
 
 func (s *StateWriter) SetInitialState() {
+	// TODO: Should we also set Anchor field here?
```
nnikolash: I assume we should also set the anchor for the first block. But the problem is that `SetInitialState()` is called from inside of `InitChain()`, which does not have anchor as an argument.
Is some cases `InitChain` is called from `InitChainByAnchor`, which does have anchor argument. But there are also couple cases when `InitChain` is called on its own, without anchor. There's also a lot of such places in tests.
So for now IDK if I should replace all calls to `InitChain` with `InitChainByAnchor` and if that makes sense. Will continue my journey.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -18,6 +19,7 @@ type BlockInfo struct {
 	SchemaVersion         uint8
 	BlockIndex            uint32
 	Timestamp             time.Time
+	Anchor                *isc.StateAnchor
```
dessaya: I think you want to store the `PreviousAnchor`. i.e. on block `n` you store the anchor `n-1`. And since the anchor includes the L1Commitment then maybe you can replace the `PreviousL1Commitment`.

packages/vm/core/blocklog/impl.go
dessaya: In block `n` we store anchor `n-1`, except for the origin block.

packages/vm/core/blocklog/blockinfo.go
nnikolash: Great point, thanks 🤝 
I fixed it jsut my renaming the field and removing `PreviousL1Commitment` field.

Just FYI that I still return same value from `ISCAnchor(stateIndex uint32)`, because its result is than used to trace the request, so I treat this function as "get anchor with was used for forming that state".

packages/vm/core/blocklog/impl.go
nnikolash: Heh, yeah, thanks, that seems logical, shame I did not figured that out myself 😄
Then leaving the field empty on initialization.

packages/webapi/services/waspevmbackend.go
lmoe: The Anchors should remain. We prune transaction history, so you can use the state by stateindex

nnikolash: After discussion, we decided that my implementation is correct - leaving it as is.

PR #421, opened 2024-12-08
feat: Use error text instead of text code

howjmay:
see https://move-book.com/move-basics/assert-and-abort.html for more details


clients/iota-go/contracts/isc/sources/request.move
```diff
@@ -9,8 +9,8 @@ module isc::request {
     use std::string::String;
     use isc::assets_bag::AssetsBag;
 
-    // The allowance coin_types vector and balances vector are not in the same size
-    const EAllowanceVecUnequal: u64 = 0;
+    #[error]
+    const EAllowanceVecUnequal: vector<u8> = b"The allowance's coin_types and balances vector are not in the same size";
```
lmoe, line 13: Do error messages get persisted - i.e do they have a byte cost?

PR #422, opened 2024-12-08
feat: Implement Ledger to wasp-cli

lmoe:
# Description of change

Adds an IOTA Rebased Ledger App implementation.
Had to copy the Zondax Library to add a Speculos testing implementation and change some minor stuff

PoC / Draft for now, but everything seems to work (GetVersion, GetPublicKey/Address, SignTransaction).
For now just tried to make it work, so keep the actual review for later. :D


clients/iota-go/hw_ledger/signer.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func NewLedgerSigner(device *HWLedger, bip32Path string, askForPublicKeyConfirmation bool) *Signer {
+	return &Signer{
+		device:                      device,
+		bip32Path:                   bip32Path,
+		askForPublicKeyConfirmation: askForPublicKeyConfirmation,
+	}
+}
+
```
nnikolash, line 23: Maybe it makes sense to check that it implements `iotasigner.Signer` interface?

clients/iota-go/hw_ledger/ledger.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+func TryAndConnect() (*HWLedger, error) {
+	l := ledgergo.NewLedgerHIDTransport()
+
+	if l.CountDevices() == 0 {
+		return nil, ErrDeviceNotFound
+	}
+
```
nnikolash, line 38: Ofc super rare case, but maye we should somehow indicate that we detect multiple ledgers? Maybe fail, or warn, or prompt? IDK

clients/iota-go/hw_ledger/ledger.go
```diff
@@ -0,0 +1,221 @@
...
+	// Write each path element as little-endian uint32
+	for i, element := range paths {
+		binary.LittleEndian.PutUint32(payload[1+4*i:], uint32(element))
+	}
+
+	return payload, nil
+}
+
+func splitPath(path string) ([]uint32, error) {
```
nnikolash, line 184: Maybe this function could benefit from having a comment with examples about what it does

clients/iota-go/hw_ledger/ledger-go/ledger_speculos.go
```diff
@@ -0,0 +1,148 @@
+package ledger_go
+
+import (
+	"encoding/binary"
+	"encoding/json"
+	"fmt"
+	"net"
+	"time"
+)
+
```
nnikolash, line 10: Maybe it would be nice to add here one line explaining that Speculos is a simulator. Or maybe not - it is used in tests anyway 🤷🏻

clients/iota-go/hw_ledger/ledger-go/ledger_speculos.go
```diff
@@ -0,0 +1,148 @@
...
+			return nil, fmt.Errorf("failed to connect to automation socket: %w", err)
+		}
+		transport.automationSocket = autoSocket
+
+		// Handle automation events in a goroutine
+		go transport.handleAutomationEvents()
+	}
+
+	// Wait a bit to ensure connection is stable
+	time.Sleep(100 * time.Millisecond)
```
nnikolash, line 62: IDK really if that might be a real issue, but I remember how some of the unit tests were failing for me, because I had slow PC and default timeout was not enough.
In this case its different and this must be enough, but maybe to be on the safe side it still makes sense to let user control this line? The changes of issue are super low, but the consequence would be very painful to the user.

On another side - it seems at user can always workaround it by simply sleeping more outside of this function 😄

clients/iota-go/hw_ledger/ledger-go/ledger_speculos.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Read response
+	sizeBuf := make([]byte, 4)
+	_, err = t.apduSocket.Read(sizeBuf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response size: %w", err)
+	}
+
+	//
+	size := binary.BigEndian.Uint32(sizeBuf)
```
nnikolash, line 101: I really dont know if this encoding is somehow related to encoding on the line 108. But I just spotted that here big endian is used while over there it could be anything. So I decided just point it out

nnikolash (approved):

PR #431, opened 2024-12-13
make solobench compilable

howjmay:


packages/solo/solobench/solobench.go
```diff
@@ -25,33 +23,34 @@ func RunBenchmarkSync(b *testing.B, chain *solo.Chain, reqs []*solo.CallParams,
...
-		requestMap := map[isc.ChainID][]isc.Request{
-			chain.ChainID: {
-				iscRequest,
-			},
-		}
-
-		go chain.Env.EnqueueRequests(requestMap)
-	}
-	require.True(b, chain.WaitForRequestsThrough(b.N, 20*time.Second))
+	panic("TODO")
```
lmoe, line 26: Can't you just Fix the call itself? :D Doesnt seem too complicated

lmoe (approved):

packages/solo/solobench/solobench.go
howjmay: I am not sure how to replace `EnqueueRequests` and `WaitForRequestsThrough`. So I just marked them with the panic here. And fix maybe later. It is quite annoying that I can't run a global go test, because of this simple error

PR #432, opened 2024-12-13
fix: Fixed couple simple tests

nnikolash:


packages/evm/evmtest/wiki_how_tos/how_tos_test.go
```diff
@@ -44,9 +44,9 @@ func TestBaseBalance(t *testing.T) {
 
 	balance, _ := env.Chain.EVM().Balance(deployer, nil)
 	decimals := env.Chain.EVM().BaseToken().Decimals
-	var value coin.Value
+	var value uint64
```
nnikolash, line 47: Unfortunatelly it seems that go-ethereum cannot marshal into "custom" type even if it is based on `uint64` and ABI also has `uint64` (check `go-ethereum@v1.14.5-wasp1/accounts/abi/reflect.go`).

lmoe: Is it completely unable to add custom types? I have seen the issue in the tests a few times and didn't fix it because I thought that it might be fixed better in another way to actually support coin.Type.

If its not possible, then lgtm :)

lmoe (approved):

packages/evm/evmtest/wiki_how_tos/how_tos_test.go
howjmay: Is there any other place may use coin.Value for goeth functions?

nnikolash: > Is it completely unable to add custom types?

Yeah, I spent some time searching it, but after brief look did not find anything - not in google nor in the code. But ok, let me do another round if you believe it would be nice to have.

nnikolash: IDK, gpt says `MarshalABI`/`UnmarshalABI` to implement, but seems he is lying 😄 because I tried and it did not work and I do not see anything about it on the internet.

I also tried copying interface of `types.Transaction`, which I understand is unrelated, but ... just in case. Obviously it did not do anything.

Finally I went through the code, debugged the branch it passes, and I do not see any possibilities for customization there.
So I would say thete is no such option.

PR #433, opened 2024-12-14
Apply golint fix

lmoe:


clients/iota-go/iotajsonrpc/types.go
```diff
@@ -9,9 +9,9 @@ import (
 	"github.com/iotaledger/wasp/clients/iota-go/iotago"
 )
 
-const (
-	IotaCoinType   = "0x2::iota::IOTA"
-	IotaCoinSupply = 5380677226936000000
+var (
+	IotaCoinType   CoinType = CoinType(iotago.MustNewResourceType("0x2::iota::IOTA").String())
+	IotaCoinSupply          = 5380677226936000000
```
howjmay: Maybe remove this? Because the supply is not static now

clients/iscmove/iscmoveclient/client_anchor.go
```diff
@@ -51,15 +51,15 @@ func (c *Client) StartNewChain(
...
+	if req.ChainGasCoin != nil {
+		ptb = PTBOptionSomeIotaCoin(ptb, req.ChainGasCoin)
 	} else {
 		ptb = PTBOptionNoneIotaCoin(ptb)
 	}
-	argInitCoin = ptb.LastCommandResultArg()
+	chainGasCoin = ptb.LastCommandResultArg()
 
-	ptb = PTBStartNewChain(ptb, req.PackageID, req.StateMetadata, argInitCoin, req.ChainOwnerAddress)
+	ptb = PTBStartNewChain(ptb, req.PackageID, req.StateMetadata, chainGasCoin, req.ChainOwnerAddress)
```
howjmay: InitCoin and GasCoin are different things. InitCoin will be put into AssetsBag, but GasCoin won't. I think this part we should not change it here, and open another PR to clarify the behavior here

clients/iscmove/iscmoveclient/client_anchor_test.go
```diff
@@ -32,7 +34,7 @@ func TestStartNewChain(t *testing.T) {
 			ChainOwnerAddress: signer.Address(),
 			PackageID:         l1starter.ISCPackageID(),
 			StateMetadata:     []byte{1, 2, 3, 4},
-			InitCoinRef:       getCoinsRes.Data[1].Ref(),
+			ChainGasCoin:      getCoinsRes.Data[1].Ref(),
```
howjmay: same here

howjmay (changes requested):

clients/iscmove/iscmoveclient/client_anchor_test.go
```diff
@@ -122,16 +126,26 @@ func TestReceiveRequestAndTransition(t *testing.T) {
...
 	anchor, err := client.StartNewChain(
 		context.Background(),
 		&iscmoveclient.StartNewChainRequest{
 			Signer:            signer,
 			ChainOwnerAddress: signer.Address(),
 			PackageID:         l1starter.ISCPackageID(),
 			StateMetadata:     []byte{1, 2, 3, 4},
-			InitCoinRef:       getCoinsRes.Data[1].Ref(),
+			ChainGasCoin:      selectedChainGasCoin.Ref(),
+			GasPayments:       []*iotago.ObjectRef{gasCoin.Ref()},
```
howjmay, line 148: same here

packages/solo/solo.go
```diff
@@ -674,12 +675,17 @@ func (env *Solo) executePTB(
...
+				ShowObjectChanges:  true,
+				ShowEvents:         true,
+				ShowInput:          true,
+				ShowBalanceChanges: true,
+				ShowRawEffects:     true,
+				ShowRawInput:       true,
 			},
 		},
 	)
+	fmt.Print(execRes.Digest.String())
```
howjmay: Not sure if this change is kept deliberately

howjmay (changes requested):

clients/iota-go/iotatest/coin.go
```diff
@@ -0,0 +1,79 @@
...
+		context.Background(),
+		iotaclient.GetCoinsRequest{Owner: cryptolibSigner.Address()},
+	)
+	require.NoError(t, err)
+
+	if len(getCoinsRes.Data) > 1 {
+		return
+	}
+
+	coins, err := client.GetCoinObjsForTargetAmount(
```
howjmay, line 32: I think this func returns multiple coins to meet the target value. But when you construct the `txData` in the following, only the first element is used

howjmay (changes requested):

clients/iota-go/iotatest/coin.go
lmoe: In my understanding it just returns multiple coin objects matching the balance, at least all pulled balances I got were over the target amount, so I just use it to select a coin

PR #434, opened 2024-12-16
fix offledger_test and ImpersonatedOffLedgerRequest

howjmay:


packages/cryptolib/signature.go
```diff
@@ -121,6 +121,9 @@ func (s *Signature) Read(r io.Reader) error {
 
 func (s *Signature) Write(w io.Writer) error {
 	ww := rwutil.NewWriter(w)
+	if *s == *NewEmptySignature() {
```
lmoe, line 124: This looks a bit odd. If the signature is empty we maybe should just serialize it in the normal form, just with zeroed bytes. Otherwise the Read function would probably throw errors on the other side.

lmoe (approved):

packages/cryptolib/signature.go
howjmay: I will solve it in the coming PR

PR #437, opened 2024-12-18
feat: Add L1Syncer to sync the L1 state

howjmay:


packages/parameters/l1parameters.go
```diff
@@ -26,95 +26,144 @@ type BaseToken struct {
...
+
+		l.epochID = newEpochID
+		// theoretically we can check once a day only, then startTimestamp will be needed
+		l.startTimestamp = l.startTimestamp.UTC()
+		l.epochDuration = time.Duration(summary.EpochDurationMs.Uint64()) * time.Millisecond
+
+		l1ParamsMutex.Lock()
+		l1Params.Protocol = summary
+		l1Params.BaseToken.TokenSupply = totalSupply.Value.Uint64()
+		l1ParamsMutex.Unlock()
```
dessaya: Mutating the object is not safe, even inside the mutex lock, because the same reference might be used by another goroutine concurrently.

packages/parameters/l1parameters.go
```diff
@@ -26,95 +26,144 @@ type BaseToken struct {
...
 }
 
-// InitL1Lazy sets a function to be called the first time L1() is called.
-// The function must call InitL1().
-func InitL1Lazy(f func()) {
-	l1ParamsLazyInit = f
+// Check latest system sttate every hour
+func (l *L1Syncer) Start() {
+	l.status = true
+	for l.status {
```
dessaya, line 219: I'm not sure updating the object asynchronously is the right approach. When the VM is executed, every bit of input data must be deterministic, including the L1 params.

Most notably:

- The L1 params must NOT change in the middle of a VM run.
- When re-executing a past transaction (e.g. for view calls or tracing) we need to use the same L1 params that were used when the block was created.

howjmay: It sounds like it is better to pass L1Params as a field in vm.VMTask. However, referenceGas could a an exception here? It is kind strange that reuse an outdated referenceGas. It may be too low to publish the tx

howjmay: If it is a view call, may be it is ok that we cant promise it use the same L1Params. We dont know how long the request will be executed.
I think most of the values in L1Params are still stable most of the time. 
The only two we are using and will change often are `ReferenceGas` and `TotalSupply`. The first one should not be used in VM and I dont see any meaningful usage of `TotalSupply` except a `function totalSupply() public view virtual returns (uint256)` in `iscmagic/ERC20Coin.sol`

dessaya: If the reference gas is used at all in the computation of the ransition PTB then yes, it has to be provided in VMTask. Otherwise maybe we can skip it.

Maybe the best approach is to store the L1Params in the blocklog, just like we store the PreviousAnchor.

dessaya: If we can guarantee that the only changing values are ReferenceGas and TotalSupply, then we can store those in the blocklog and not the rest.

And we have to ensure that any L2 logic that depends on those values read them from the blocklog.

howjmay: Good idea. Let me work on it

clients/iota-go/iotajsonrpc/bigint.go
```diff
@@ -50,3 +50,12 @@ func (w *BigInt) UnmarshalJSON(data []byte) error {
 func (w *BigInt) MarshalJSON() ([]byte, error) {
 	return json.Marshal(w.String())
 }
+
+func (w *BigInt) Clone() *BigInt {
+	ret := NewBigInt(0)
+	err := ret.UnmarshalText([]byte(w.String()))
+	if err != nil {
+		panic(err)
+	}
```
dessaya: More efficient: `ret.Set(w.Int)`

packages/chain/chain.go
```diff
@@ -92,6 +92,7 @@ type ChainCore interface {
 	LatestAnchor(freshness StateFreshness) (*isc.StateAnchor, error)
 	LatestState(freshness StateFreshness) (state.State, error)
 	LatestGasCoin(freshness StateFreshness) (*coin.CoinWithRef, error)
+	LatestL1Parameter() *parameters.L1Params
```
dessaya: Nit: should be `LatestL1Parameters` or `LatestL1Params`

packages/chainutil/evmcall.go
```diff
@@ -18,6 +18,7 @@ import (
 // EVMCall executes an EVM contract call and returns its output, discarding any state changes
 func EVMCall(
 	anchor *isc.StateAnchor,
+	l1Param *parameters.L1Params,
```
dessaya: Same everywhere, `l1Params` to keep consistency

packages/isc/coininfo.go
```diff
@@ -13,7 +13,8 @@ type IotaCoinInfo struct {
 	Symbol      string
 	Description string
 	IconURL     string
-	TotalSupply coin.Value
+	// TotalSupply is a dynamic value
+	// TotalSupply coin.Value
```
dessaya: Why is this removed? The total supply is necessary for ERC20 contracts in EVM. The idea is to keep this updated in the accounts contract.

packages/parameters/l1parameters.go
```diff
@@ -26,95 +80,158 @@ type BaseToken struct {
 	Subunit         string    `json:"subunit" swagger:"desc(The token subunit),required"`
 	Decimals        uint8     `json:"decimals" swagger:"desc(The token decimals),required"`
 	UseMetricPrefix bool      `json:"useMetricPrefix" swagger:"desc(Whether or not the token uses a metric prefix),required"`
-	CoinType        coin.Type `json:"coinType"`
+	CoinType        coin.Type `json:"coinType" swagger:"desc(BaseToken's Cointype),required"`
+	TokenSupply     uint64    `json:"tokenSupply" swagger:"desc(The total supply of BaseToken),required"`
```
dessaya: should be `TotalSupply`?

packages/parameters/l1parameters.go
```diff
@@ -26,95 +80,158 @@ type BaseToken struct {
...
-	l1ParamsLazyInit func()
 )
 
-func isTestContext() bool {
-	return strings.HasSuffix(os.Args[0], ".test") ||
-		strings.HasSuffix(os.Args[0], ".test.exe") ||
-		strings.Contains(os.Args[0], "__debug_bin")
+func L1() *L1Params {
+	l1ParamsMutex.RLock()
+	ret := l1Params.Clone()
```
dessaya: If the object is immutable then cloning should not be necessary here.

packages/webapi/services/waspevmbackend.go
```diff
@@ -79,6 +79,7 @@ func (b *WaspEVMBackend) EVMSendTransaction(tx *types.Transaction) error {
 func (b *WaspEVMBackend) EVMCall(anchor *isc.StateAnchor, callMsg ethereum.CallMsg) ([]byte, error) {
 	return chainutil.EVMCall(
 		anchor,
+		b.chain.LatestL1Parameter(),
```
dessaya: When making view calls and tracing, it would be incorrect to use the latest L1Params. It should be the one that was in effect when the given anchor was transitioned.

We are storing the L1Params in the blocklog for this reason, no?

packages/solo/solo.go
```diff
@@ -159,6 +160,15 @@ func New(t Context, initOptions ...*InitOptions) *Solo {
 		}
 	}
 
+	l1Syncer := parameters.NewL1Syncer(
+		iotaclient.NewHTTP(l1starter.Instance().APIURL()),
+		10*60*time.Second,
+		opt.Log,
+	)
+	go l1Syncer.Start()
+	time.Sleep(200 * time.Millisecond)
```
dessaya: What is this sleep for?

packages/vm/core/blocklog/impl.go
```diff
@@ -26,6 +27,7 @@ func (s *StateWriter) SetInitialState() {
 		SchemaVersion:         BlockInfoLatestSchemaVersion,
 		BlockIndex:            0,
 		Timestamp:             time.Time{},
+		L1Params:              parameters.L1Default,
```
dessaya: This should probably be the actual L1 params at the time of chain creation?

tools/wasp-cli/cli/setup/init.go
```diff
@@ -28,13 +31,7 @@ func Init(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(initRefreshL1ParamsCmd())
 
 	// The first time parameters.L1() is called, it will be initialized with this function
```
dessaya: This comment is no longer relevant

tools/wasp-cli/cli/setup/init.go
```diff
@@ -28,13 +31,7 @@ func Init(rootCmd *cobra.Command) {
...
-
-		if config.L1ParamsExpired() {
-			config.RefreshL1ParamsFromNode()
-		} else {
-			config.LoadL1ParamsFromConfig()
-		}
-	})
+	client := iotaclient.NewHTTP(config.L1APIAddress())
+	l1Syncer := parameters.NewL1Syncer(client, 600*time.Second, log.HiveLogger())
+	go l1Syncer.Start()
```
dessaya: Spawning a goroutine for the wasp-cli seems overkill, since it is usually a short-lived process.

dessaya (commented):
Honestly I'm not sure that having a goroutine in the background is the best approach, since having the L1 params updated at random moments seems prone to subtle bugs.

Solo and wasp-cli don't need it, only the consensus. So maybe the consensus should be the one responsible for updating the L1 params whenever it sees fit to do so.

packages/solo/solo.go
howjmay: I have change this to update only once here

packages/vm/core/blocklog/impl.go
howjmay: This is the initial state I think it should be ok to use the default one. I think the VM will not use this value for process any incoming requests?
The usage of `SetInitialState` are the func which dont have access of L1. I tried a little bit already, but it will change quite a lot of things. So maybe we can keep it `parameters.L1Default`?

packages/parameters/l1parameters.go
```diff
@@ -26,95 +80,159 @@ type BaseToken struct {
...
-
-	l1ParamsLazyInit func()
 )
 
-func isTestContext() bool {
-	return strings.HasSuffix(os.Args[0], ".test") ||
-		strings.HasSuffix(os.Args[0], ".test.exe") ||
-		strings.Contains(os.Args[0], "__debug_bin")
+func L1() *L1Params {
+	return l1Params
```
dessaya, line 141: What happened to the mutex? I think it's needed here, since reading the pointer is not guaranteed to be atomic.

packages/parameters/l1parameters.go
```diff
@@ -26,95 +80,159 @@ type BaseToken struct {
...
-		},
-		BaseToken: &BaseToken{
-			Name:            "Iota",
-			TickerSymbol:    "MIOTA",
-			Unit:            "MIOTA",
-			Subunit:         "IOTA",
-			Decimals:        6,
-			UseMetricPrefix: false,
+	// FIXME this should set to init in the beginning, otherwise, L2 may use outdated data to calculate
+	l1Params *L1Params
```
dessaya: In the previous version this is intentionally left unset, because using outdated params is incorrect. This forced other parts of the code to call `Init` first with the most recent version of the params.

There was even a check:

```
if l1Params == nil {
		panic("call InitL1() first")
}
```

I think this check is important, so that if someone forgets to call `InitL1` it panics immediately with a relevant error message, instead of just blowing up in a random place with a nil pointer access.

So, either initialize with `L1Default`, or add the nil check.

packages/solo/solo.go
```diff
@@ -159,6 +160,13 @@ func New(t Context, initOptions ...*InitOptions) *Solo {
 		}
 	}
 
+	l1Syncer := parameters.NewL1Syncer(
+		iotaclient.NewHTTP(l1starter.Instance().APIURL()),
+		10*60*time.Second,
+		opt.Log,
+	)
+	l1Syncer.Upate()
```
dessaya: Maybe add a `parameters.InitL1(client, log)` function?

packages/vm/core/blocklog/impl.go
dessaya: Maybe it's ok for now, but I think it's better to always use the most recent L1Params, to avoid subtle bugs.
Why is it difficult to change? I would just add it as a parameter in `SetInitialState()`. At least add a TODO?

packages/vm/core/evm/iscmagic/types.go
```diff
@@ -252,6 +253,7 @@ func WrapIotaCoinInfo(info *isc.IotaCoinInfo) IotaCoinInfo {
 		Symbol:      info.Symbol,
 		Description: info.Description,
 		IconUrl:     info.IconURL,
-		TotalSupply: CoinValue(info.TotalSupply),
+		// FIXME this will cause panic
```
dessaya: Is this still true?

packages/webapi/services/waspevmbackend.go
```diff
@@ -180,3 +198,20 @@ func (*WaspEVMBackend) RevertToSnapshot(int) error {
...
+	store := b.chain.Store()
+	stateMetadata, err := transaction.StateMetadataFromBytes(anchor.GetStateMetadata())
+	if err != nil {
+		return nil, err
+	}
+	state, err := store.StateByTrieRoot(stateMetadata.L1Commitment.TrieRoot())
+	if err != nil {
+		return nil, err
+	}
+	_, blockInfo, err := corecontracts.GetBlockInfo(b.chain, state.BlockIndex(), stateMetadata.L1Commitment.TrieRoot().String())
```
dessaya: Avoid making view calls here. Instead, you can access the state directly using `blocklog.NewStateReaderFromChainState()` which is far more efficient.

```go
func (b *WaspEVMBackend) getBlockInfoByAnchor(anchor *isc.StateAnchor) (*blocklog.BlockInfo, error) {
	stateMetadata, err := transaction.StateMetadataFromBytes(anchor.GetStateMetadata())
	if err != nil {
		return nil, err
	}
	state, err := b.ISCStateByTrieRoot(stateMetadata.L1Commitment.TrieRoot())
	if err != nil {
		return nil, err
	}
	blockInfo, ok := blocklog.NewStateReaderFromChainState(state).GetBlockInfo(anchor.GetStateIndex())
	if !ok {
		return nil, fmt.Errorf("blockinfo not found")
	}
	return blockInfo, nil
}
```

packages/solo/evm.go
```diff
@@ -56,6 +56,7 @@ func (b *jsonRPCSoloBackend) EVMSendTransaction(tx *types.Transaction) error {
 func (b *jsonRPCSoloBackend) EVMCall(anchor *isc.StateAnchor, callMsg ethereum.CallMsg) ([]byte, error) {
 	return chainutil.EVMCall(
 		anchor,
+		b.Chain.LatestL1Parameters(),
```
dessaya: In these functions it is also necessary to use the correct L1Params, instead of the latest one.

I think it's better to add L1Params as a parameter in `EVMCall`, `EVMEstimateGas` & `EVMTraceTransaction`.

packages/vm/core/blocklog/impl.go
howjmay: `SetInitialState ()` is used by `InitChain()`. And Most of the places using `InitChain()` doesnt have access to L1Client. Cyclic import happened when I tried to fetch the latest L1params here
I can leave a TODO here.

packages/parameters/l1parameters.go
howjmay: I have initialized to L1Default

packages/vm/core/evm/iscmagic/types.go
howjmay: fixed because of initialization

packages/webapi/services/waspevmbackend.go
howjmay: fixed

packages/solo/evm.go
howjmay: added

packages/solo/solo.go
howjmay: added

packages/vm/core/evm/iscmagic/types.go
```diff
@@ -252,6 +253,6 @@ func WrapIotaCoinInfo(info *isc.IotaCoinInfo) IotaCoinInfo {
 		Symbol:      info.Symbol,
 		Description: info.Description,
 		IconUrl:     info.IconURL,
-		TotalSupply: CoinValue(info.TotalSupply),
+		TotalSupply: CoinValue(parameters.L1().BaseToken.TotalSupply),
```
dessaya: IMO it's better to leave it as it was (`info.TotalSupply`) to avoid calling the global function.

packages/webapi/services/waspevmbackend.go
```diff
@@ -180,3 +186,19 @@ func (*WaspEVMBackend) RevertToSnapshot(int) error {
 func (*WaspEVMBackend) TakeSnapshot() (int, error) {
 	return 0, errNotImplemented
 }
+
+func (b *WaspEVMBackend) getBlockInfoByAnchor(anchor *isc.StateAnchor) (*blocklog.BlockInfo, error) {
```
dessaya: This is unused now?

tools/wasp-cli/cli/setup/init.go
```diff
@@ -27,14 +30,7 @@ func Init(rootCmd *cobra.Command) {
...
-
-		if config.L1ParamsExpired() {
-			config.RefreshL1ParamsFromNode()
-		} else {
-			config.LoadL1ParamsFromConfig()
-		}
-	})
+	client := iotaclient.NewHTTP(config.L1APIAddress(), iotaclient.WaitForEffectsDisabled)
+	l1Syncer := parameters.NewL1Syncer(client, 600*time.Second, log.HiveLogger())
+	l1Syncer.Upate()
```
dessaya:
```suggestion
	l1Syncer.InitL1(...)
```

packages/parameters/l1parameters.go
```diff
@@ -26,95 +80,203 @@ type BaseToken struct {
...
+	// theoretically we can check once a day only, then startTimestamp will be needed
+	l.startTimestamp = l.startTimestamp.UTC()
+	l.epochDuration = time.Duration(protocol.EpochDurationMs.Uint64()) * time.Millisecond
+
+	l1ParamsMutex.Lock()
+	newL1Params := l1Params.Clone()
+	newL1Params.Protocol = &protocol
+	newL1Params.BaseToken.TotalSupply = totalSupply.Value.Uint64()
+	l1Params = newL1Params
+	l1ParamsMutex.Unlock()
```
dessaya, line 277: There is duplicated code in `InitL1` and `Update`.

howjmay: They are not. `InitL1()` doesnt check EpochID. The Epoch change once a day. So in theory `Update` will update L1Param once a day only. I let it sleep for an hour is in case the node has error. We can retry for every an hour in this way

dessaya: It think it is still possible (and worth it) to try and avoid the duplicated code, even if both functions are not identical. They seem mostly identical, except for this difference with the epoch.

dessaya (approved):

howjmay:
I think I can remove L1Syncer in the following PR. So let me merge it first

PR #441, opened 2024-12-29
fix: Fix linting warnings

howjmay:


clients/iota-go/hw_ledger/ledger-go/apduWrapper.go
```diff
@@ -112,7 +113,7 @@ func SerializePacket(
 	codec.PutUint16(buffer[3:], sequenceIdx)
 
 	// Only insert total size of the command in the first package
-	if sequenceIdx == 0 {
+	if sequenceIdx == 0 && len(command) < math.MaxUint16 {
```
dessaya: This is incorrect IMO, because If `len(command) >= math.MaxUint16` then the serialization will be incorrect.

Instead:

```suggestion
	if len(command) > math.MaxUint16 {
		panic(...)
	}
	if sequenceIdx == 0 {
```

dessaya (approved):

PR #445, opened 2025-01-07
feat: Check transaction size

howjmay:
In `ProtocolConfig` there are two more const will affect us. They are
* max_size_written_objects
* max_serialized_tx_effects_size_bytes

The problem is we can't know how many object we have written (mutated, created, unwrapped) during execution, because we dont know the size of AssetBag.
And the `max_serialized_tx_effects_size_bytes` regulates the size of Move execution result, which is impossible for us to get too.

### DIFF
--- a/packages/vm/vmimpl/runreq.go
+++ b/packages/vm/vmimpl/runreq.go
@@ -56,7 +56,7 @@ func (vmctx *vmContext) runRequest(req isc.Request, requestIndex uint16, mainten
 
 	result, err := reqctx.callTheContract()
 	if err == nil {
-		err = vmctx.checkTransactionSize()
+		err = vmctx.txbuilder.CheckTransactionSize()
 	}
 	if err != nil {
 		// skip the request / rollback tx builder (no need to rollback the state, because the mutations will never be applied)
@@ -460,15 +460,4 @@ func (reqctx *requestContext) GetContractRecord(contractHname isc.Hname) (ret *r
 
 func (vmctx *vmContext) loadChainConfig() {
 	vmctx.chainInfo = governance.NewStateReaderFromChainState(vmctx.stateDraft).GetChainInfo(vmctx.ChainID())
-}
-
-// checkTransactionSize panics with ErrMaxTransactionSizeExceeded if the estimated transaction size exceeds the limit
-func (vmctx *vmContext) checkTransactionSize() error {
-	// TODO the following requirements we need to check
-	// * The encoded bytes of ptb should be smaller than `max_tx_size_bytes`
-	// * max_size_written_objects
-	// * max_serialized_tx_effects_size_bytes
-	// * max_pure_argument_size
-	vmctx.task.Log.Info("TODO: checkTransactionSize")
-	return nil
 }
--- a/packages/vm/vmtxbuilder/txbuilder.go
+++ b/packages/vm/vmtxbuilder/txbuilder.go
@@ -1,6 +1,7 @@
 package vmtxbuilder
 
 import (
+	"fmt"
 	"slices"
 
 	"github.com/samber/lo"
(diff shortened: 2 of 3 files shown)


packages/vm/vmimpl/runreq.go
```diff
@@ -477,11 +481,23 @@ func (vmctx *vmContext) loadChainConfig() {
...
-	// * max_size_written_objects
-	// * max_serialized_tx_effects_size_bytes
-	// * max_pure_argument_size
-	vmctx.task.Log.Info("TODO: checkTransactionSize")
+	const maxTxSizeBytes = 128 * 1024
+	const maxInputObjects = 2048
+	const maxProgrammableTxCommands = 1024
+	ptb := vmctx.txbuilder.ViewPTB()
+	if ptb.Inputs.Len() > maxInputObjects {
+		return fmt.Errorf("ptb input len: %d, exceed max_input_objects", ptb.Inputs.Len())
```
dessaya: The returned error must be `vmexceptions.ErrMaxTransactionSizeExceeded`.

packages/vm/vmimpl/runreq.go
```diff
@@ -477,11 +481,23 @@ func (vmctx *vmContext) loadChainConfig() {
 
 // checkTransactionSize panics with ErrMaxTransactionSizeExceeded if the estimated transaction size exceeds the limit
```
dessaya:
```suggestion
// checkTransactionSize returns ErrMaxTransactionSizeExceeded if the estimated transaction size exceeds the limit
```

packages/vm/vmimpl/runreq.go
```diff
@@ -477,11 +481,23 @@ func (vmctx *vmContext) loadChainConfig() {
...
 func (vmctx *vmContext) checkTransactionSize() error {
-	// TODO the following requirements we need to check
-	// * The encoded bytes of ptb should be smaller than `max_tx_size_bytes`
-	// * max_size_written_objects
-	// * max_serialized_tx_effects_size_bytes
-	// * max_pure_argument_size
-	vmctx.task.Log.Info("TODO: checkTransactionSize")
+	const maxTxSizeBytes = 128 * 1024
+	const maxInputObjects = 2048
+	const maxProgrammableTxCommands = 1024
```
dessaya: These numbers are arbitrary?

packages/vm/vmtxbuilder/txbuilder.go
```diff
@@ -145,3 +145,7 @@ func (txb *AnchorTransactionBuilder) BuildTransactionEssence(stateMetadata []byt
 	}
 	return ptb.Finish()
 }
+
+func (txb *AnchorTransactionBuilder) ViewPTB() *iotago.ProgrammableTransactionBuilder {
+	return txb.ptb.Clone()
```
dessaya, line 152: There is a lot of stuff that is added to the PTB when `BuildTransactionEssence` is called. That is not being included here? In that case the estimation will be always lower than the real size.

Look at the stardast implementation:

https://github.com/iotaledger/wasp/blob/06b7a38d2e1360fc4f9a5fadf4288383121e0b09/packages/vm/vmimpl/runreq.go#L514-L521

packages/vm/vmimpl/runreq.go
howjmay: no they are values from iota repo

howjmay: The constants here come from `struct ProtocolConfig` in `crates/iota-protocol-config/src/lib.rs`

packages/vm/vmtxbuilder/txbuilder.go
howjmay: I reformed the way to check the tx size. I build the tx every time the check func is called

dessaya (approved):

PR #448, opened 2025-01-09
Demonstration of issues for discussion

nnikolash:


packages/chain/cmt_log/var_cons_insts.go
```diff
@@ -48,7 +48,7 @@ func NewVarConsInsts(
 		},
 		minLI:     minLI,
 		maxLI:     minLI,
-		lastLI:    NilLogIndex(),
+		lastLI:    minLI,
```
nnikolash: Having `lastLI: NilLogIndex()` results in a panic, because upon call to `VarConsInsts.LatestL1AO()` value `vci.lastLI` is passed to `trySet()`, which do not pass first `if` and so `vci.outputCB()`  never called and anchor never set into map `lis`. So map `lis` continues to have default value for `minLI` entry - `nil`. Because of that `cmtLogImpl.Output()` returns map with entry with nil anchor pointer, which later used in `TrackAliasOutput()` and panics.

What is the correct way of fixing it? I do not fully understand the logic, but I assume the value was set to `NilLogIndex()` for a reason?

packages/chain/cmt_log/var_cons_insts.go
```diff
@@ -98,7 +98,7 @@ func (vci *varConsInstsImpl) trySet(li LogIndex, ao *isc.StateAnchor, cb onLIInc
 	}
 	//
 	// Is it already proposed?
-	if _, ok := vci.lis[li]; ok {
+	if existing := vci.lis[li]; existing != nil {
 		return nil
 	}
```
nnikolash: For `minLI` is set from the start - `nil`. Because of that this check does not pass. But maybe it shouldn't? Maybe I'm missing the fact, that `minLI` should neven ever be checked here and having `nil` for it is correct?

packages/chain/node.go
```diff
@@ -675,7 +675,7 @@ func (cni *chainNodeImpl) handleTxPublished(ctx context.Context, txPubResult *tx
 func (cni *chainNodeImpl) handleAliasOutput(ctx context.Context, aliasOutput isc.StateAnchor) {
 	cni.log.Debugf("handleAliasOutput: %v", aliasOutput)
 	if aliasOutput.GetStateIndex() == 0 {
-		initBlock, err := origin.InitChainByAnchor(cni.chainStore, &aliasOutput, 0, isc.BaseTokenCoinInfo)
+		initBlock, err := origin.InitChainByAnchor(cni.chainStore, &aliasOutput, 2000000000, isc.BaseTokenCoinInfo)
```
nnikolash: This is just a temporary fix for mismatch of L1Commitments.

packages/chain/node_test.go
```diff
@@ -150,19 +150,24 @@ func testNodeBasic(t *testing.T, n, f int, reliable bool, timeout time.Duration,
...
 			require.NoError(t, err)
-			collectedRequests = append(collectedRequests, onLedgerRequest)
 			tnc.recvRequest(
 				onLedgerRequest,
 			)
 		}
 	}
 
-	awaitRequestsProcessed(ctxTimeout, te, collectedRequests, "incRequests")
+	<-waitCh
```
nnikolash, line 170: I restructured a code, so that first we set all waiters, and only then start processing requests. This is to avoid races. Maybe it is redundant - need to discuss.

packages/chain/node_test.go
```diff
@@ -150,19 +150,24 @@ func testNodeBasic(t *testing.T, n, f int, reliable bool, timeout time.Duration,
 		}
 	}
 
-	collectedRequests := make([]isc.Request, 0)
+	reqIDs := make([]isc.RequestID, 0)
+	for _, req := range incRequests {
+		reqIDs = append(reqIDs, isc.RequestID(*req.ObjectRef.ObjectID))
+	}
```
nnikolash, line 156: If I understood correctly, code was waiting for the same request multiple times, because waiting happens by ID, not by ID+nodeConn. So I changed it as part of change to waiting logic (see comment to line 170).

packages/chain/node_test.go
```diff
@@ -223,30 +228,39 @@ func testNodeBasic(t *testing.T, n, f int, reliable bool, timeout time.Duration,
...
+		for reqNum, reqID := range requestIDs {
+			te.log.Debugf("Going to AwaitRequestProcessed %v at node=%v, req[%v]=%v...", desc, i, reqNum, reqID.String())
+
+			startWaiting := func(confirmed bool) {
+				waitCh := node.AwaitRequestProcessed(ctx, reqID, confirmed)
+
+				go func() {
+					select {
+					case rec := <-waitCh:
+						if rec != nil && rec.Error != nil {
```
nnikolash, line 244: This function changed not only because I've changed logic of waiting (see comment to line 170), but I also added this `rec != nil`. This is because as far as I see from the code not each request have receipt? Or should it?... Please see comment to await_receipt.go:75.

Anyway for out test it is not needed, so I've added this command.

packages/chain/await_receipt.go
```diff
@@ -72,9 +72,6 @@ func (ari *awaitReceiptImpl) respondByState(state state.State) {
 		if err != nil {
 			panic(fmt.Errorf("cannot read recept from state: %w", err))
 		}
-		if receipt == nil {
-			return true
-		}
```
nnikolash, line 77: Deleted this, because receipt was `nil` for case of the test and I see that function still returns `true`, but does not call `Respond`. I also see, that `respondByBlock` does not have such check.
So I don't know if it is correct to delete it, need to discuss.

PR #450, opened 2025-01-11
refactor: Pass L1parameter to consensus

howjmay:


packages/chain/cons/bp/batch_proposal_set.go
```diff
@@ -141,20 +142,20 @@ func (bps batchProposalSet) selectedFeeDestination(aggregatedTime time.Time, ran
...
+func (bps batchProposalSet) aggregatedL1Params(f int) *parameters.L1Params {
+	ps := make([]*parameters.L1Params, 0, len(bps))
 	for _, bp := range bps {
-		proposals = append(proposals, bp.gasPrice)
+		ps = append(ps, bp.l1params)
 	}
-	sort.Slice(proposals, func(i, j int) bool {
-		return proposals[i] < proposals[j]
+	sort.Slice(ps, func(i, j int) bool {
+		return ps[i].Protocol.Epoch.Cmp(ps[j].Protocol.Epoch.BigInt()) < 0
```
kape1395: Byzantine nodes can send the same epoch but different parameters.
We should sort additionally on other parameters as well?
And probably we have to calculate the quorums for each parameter individually, or consider strictly equal param sets.

packages/chain/cons/cons.go
```diff
@@ -459,13 +459,13 @@ func (c *consImpl) uponSMSaveProducedBlockDone(block state.Block) gpa.OutMessage
 // NC
 
 func (c *consImpl) uponNCInputsReady(anchor *isc.StateAnchor) gpa.OutMessages {
-	c.output.NeedNodeConnGasInfo = anchor
+	c.output.NeedNodeConnL1Info = anchor
```
kape1395, line 427: Maybe a boolean would be enough?

packages/chain/cons/input_gas_info.go
dessaya, line 1: Maybe rename the file?

packages/nodeconn/nodeconn.go
```diff
@@ -178,15 +178,20 @@ func (nc *nodeConnection) ConsensusGasPriceProposal(
...
+
 		gasCoinRef := gasCoinGetObjectRes.Data.Ref()
-		var coinInfo cons_gr.NodeConnGasInfo = &SingleGasCoinInfo{
+		var coinInfo cons_gr.NodeConnL1Info = &SingleL1Info{
 			coin.CoinWithRef{
 				Type:  coin.BaseTokenType,
 				Value: coin.Value(gasCoin.Balance),
 				Ref:   &gasCoinRef,
 			},
 			// FIXME we will pass l1param from consensus
```
dessaya: Is this comment still relevant?

packages/webapi/services/waspevmbackend.go
```diff
@@ -184,3 +186,19 @@ func (*WaspEVMBackend) RevertToSnapshot(int) error {
 func (*WaspEVMBackend) TakeSnapshot() (int, error) {
 	return 0, errNotImplemented
 }
+
+func (b *WaspEVMBackend) getBlockInfoByAnchor(anchor *isc.StateAnchor) (*blocklog.BlockInfo, error) {
```
dessaya: Is this function unused?

packages/chain/cons/bp/batch_proposal_set.go
howjmay: I have changed this into returning the L1Params which shared more than `proposalCount-f` batchProposals.

packages/chain/cons/cons.go
howjmay: the anchor will be used in `ConsensusL1InfoProposal()`, so I think I have to keep it

packages/chain/cons/input_gas_info.go
howjmay: done

packages/webapi/services/waspevmbackend.go
howjmay: removed

packages/parameters/l1parameters.go
```diff
@@ -21,12 +21,17 @@ type L1Params struct {
...
-	ret += fmt.Sprintf("\tBaseToken: %s\n", l.BaseToken.String())
-	ret += "}\n"
-	return ret
+	b, _ := json.Marshal(l)
+	return string(b)
+}
+
+func L1ParamsFromString(s string) *L1Params {
+	var l L1Params
+	err := json.Unmarshal([]byte(s), &l)
```
dessaya: why is this serialized as json and not bcs?

packages/parameters/l1parameters.go
```diff
@@ -21,12 +21,17 @@ type L1Params struct {
...
+	if err != nil {
+		panic(err)
```
dessaya, line 35: also shouldn't the error be returned?

packages/parameters/l1parameters.go
howjmay: I will need it in the following Byzantine fault impl. To avoid make two different key for the same L1param value. I need to marshal them into string instead of slice.

```go
func (bps batchProposalSet) aggregatedL1Params(f int) *parameters.L1Params {
	proposalCount := len(bps) // |acsProposals| >= N-F by ACS logic.
	ps := make([]*parameters.L1Params, 0, len(bps))
	for _, bp := range bps {
		ps = append(ps, bp.l1params)
	}

	// count the amount of each L1Params
	protocolMap := make(map[string]int)
	for _, l1params := range ps {
		protocolMap[l1params.String()] += 1
	}
	for l1param, cnt := range protocolMap {
		if cnt >= proposalCount-f {
			return parameters.L1ParamsFromString(l1param)
		}
	}

	return nil
}
```

packages/parameters/l1parameters.go
howjmay: added

howjmay: In the meantime, applied the change in the above Byzantine check

packages/parameters/l1parameters.go
dessaya: If you want to use the serialized version as a key in a map, json is not applicable, since it is not deterministic. BCS is.

In Go, a `string` is just an immutable sequence of bytes. You can just convert the bytes to string as `string(bcsBytes)` and use that as a key.

dessaya: You could also calculate a hash to use as key.

howjmay: Add Hash()

kape1395:
It doesn't even compile. Have you tried to run the tests?

packages/chain/cons/bp/batch_proposal_set.go
```diff
@@ -139,18 +141,56 @@ func (bps batchProposalSet) selectedFeeDestination(aggregatedTime time.Time, ran
...
+type l1paramsCounter struct {
+	counter  int
+	l1params *parameters.L1Params
+}
+
 // Use the same aggregation logic as for the timestamp:
 // Take highest value proposed by at least F+1 nodes.
-func (bps batchProposalSet) aggregatedGasPrice(f int) uint64 {
-	proposals := make([]uint64, 0, len(bps))
+func (bps batchProposalSet) aggregatedL1Params(f int) *parameters.L1Params {
```
howjmay, line 150: @kape1395 here I do the Byzantine check for L1Params and the gas price in it? Is it necessary? Or I only need the L1Params which is shared by more than N-F nodes?

kape1395: The consensus will pick N-F proposals. F of them might be byzantine. So, you should use the value proposed by N-2F ~= F+1 nodes.

howjmay: I changed the logic here into choose the L1Params shows up the most, and it should shared by at least f+1 nodes. 
However, can I use its gas price as the aggregated gas price? Or I need to do another Byzantine check again like what I am doing here?

packages/coin/coin.go
```diff
@@ -81,6 +82,10 @@ func MustTypeFromString(s string) Type {
 	return t
 }
 
+func (t *Type) MarshalJSON() ([]byte, error) {
+	return []byte("\"" + t.String() + "\""), nil
```
dessaya: Better:

```suggestion
	return json.Marshal(t.s)
```

packages/parameters/l1parameters.go
```diff
@@ -2,63 +2,64 @@ package parameters
...
-	ret += fmt.Sprintf("\tEpoch: %s\n", p.Epoch)
-	ret += fmt.Sprintf("\tProtocolVersion: %s\n", p.ProtocolVersion)
-	ret += fmt.Sprintf("\tSystemStateVersion: %s\n", p.SystemStateVersion)
-	ret += fmt.Sprintf("\tIotaTotalSupply: %s\n", p.IotaTotalSupply)
-	ret += fmt.Sprintf("\tReferenceGasPrice: %s\n", p.ReferenceGasPrice)
-	ret += fmt.Sprintf("\tEpochStartTimestampMs: %s\n", p.EpochStartTimestampMs)
-	ret += fmt.Sprintf("\tEpochDurationMs: %s\n", p.EpochDurationMs)
-	ret += "}\n"
-	return ret
+	b, _ := json.Marshal(p)
```
dessaya: Use `lo.Must()` or `panic()` instead of ignoring errors.

Also, if this is intended as a human-readable string representation, maybe `json.MarshalIndent` is better.

packages/coin/coin.go
howjmay: done

packages/parameters/l1parameters.go
howjmay: done

packages/chain/cons/bp/batch_proposal_set.go
howjmay: @kape1395 is the new implementation is valid one? Or the logic is wrong?

packages/chain/cons/bp/batch_proposal_set.go
```diff
@@ -139,18 +141,61 @@ func (bps batchProposalSet) selectedFeeDestination(aggregatedTime time.Time, ran
...
+			}
+			protocolMap[l1params.Hash().Hex()] = elt
+		}
+		if elt.counter > l1paramsCounterMax.counter {
+			l1paramsCounterMax.counter = elt.counter
+			l1paramsCounterMax.l1params = elt.l1params
+		}
 	}
+
+	if l1paramsCounterMax.counter < f {
```
kape1395: Here should be `<=` and not `<`.

packages/chain/cons/bp/batch_proposal_set.go
```diff
@@ -139,18 +141,61 @@ func (bps batchProposalSet) selectedFeeDestination(aggregatedTime time.Time, ran
...
+		return nil
+	}
+
+	aggregatedGasPrice := aggregatedGasPrice(gasPriceBPs, f)
```
kape1395: Let's take it from `l1paramsCounterMax` instead of calculating another quorum.

packages/chain/cons/bp/batch_proposal_set.go
```diff
@@ -139,18 +141,61 @@ func (bps batchProposalSet) selectedFeeDestination(aggregatedTime time.Time, ran
...
+				l1params: l1params,
+			}
+			protocolMap[l1params.Hash().Hex()] = elt
+		}
+		if elt.counter > l1paramsCounterMax.counter {
+			l1paramsCounterMax.counter = elt.counter
+			l1paramsCounterMax.l1params = elt.l1params
+		}
 	}
+
```
kape1395, line 177: Please add a condition that if more than 1 L1params has counter > f, then NIL.
Otherwise, nodes can aggregate different values.

packages/chain/cons/bp/batch_proposal_set.go
kape1395: Fixed.

packages/chain/cons/bp/batch_proposal_set.go
kape1395: Done that myself.

kape1395 (approved):

packages/chain/cons/bp/batch_proposal_set.go
howjmay: I added it too

PR #451, opened 2025-01-11
Regenerate client with openapi-generator 7.10.0, fix naming differences

lmoe:
# Description of change
Our uint32 fork of openapi-generator was getting quite old. It was merged to the current state. 
https://github.com/lmoe/openapi-generator/tree/feature/go_gen_unsigned_int

The fork has been pushed to DockerHub
https://hub.docker.com/layers/lukasmoe/openapi-generator/latest/images/sha256-168c45f6fc3b22454e497e502e1726d59aef8eb744aa2e8008a564062cc5cd2c

The generator now adds a custom UnmarshalJSON function to each API call. We will need to determine if it works as expected for us, or if we need to disable this custom unmarshaller again. 

Previously, the main routes were called `RouteApi`. Now it's called `RouteAPI`  so all references to the API clients have been refactored.


clients/apiclient/.openapi-generator/FILES
```diff
@@ -18,7 +18,7 @@ docs/AddUserRequest.md
 docs/AliasOutputMetricItem.md
 docs/AssetsJSON.md
 docs/AssetsResponse.md
-docs/AuthApi.md
+docs/AuthAPI.md
```
howjmay, line 21: I am not familiar to openapi generator. I saw the files' names havent bee changed. They are still `docs/AuthApi.md`. Not sure if it is sth we should change

lmoe: Ah, good catch. It probably didn't update those because I made the mistake of using Windows for once, where file names are case insensitive. :D

howjmay: Should we change this?

lmoe: Yup. I was trying to fix it, but it didn't work. Seems like even with WSL (The Linux system on Windows) it has the same kind of issue.

I regenerated everything on my Linux notebook. Now the files seem to look fine.

howjmay: Sorry I just wasnt sure whether I should review or not. 
I saw the markdown files under apiclients are still named as `*Api.md`. Should we change them?

lmoe: Probably not. We generate two clients. Go and TypeScript. The Go docs land in ./docs and the TypeScript ones in ./ I assume only the Go generator changed the way of naming. I consider it mergeable.
Might be worth to move the TypeScript client into a subfolder some time however.

howjmay (approved):

PR #453, opened 2025-01-13
chore: Removed origin deposit

nnikolash:


packages/origin/origin.go
```diff
@@ -143,7 +137,7 @@ func InitChain(
 
 	// init the state of each core contract
 	root.NewStateWriter(root.Contract.StateSubrealm(d)).SetInitialState(v, contracts)
-	accounts.NewStateWriter(v, accounts.Contract.StateSubrealm(d)).SetInitialState(originDeposit, baseTokenCoinInfo)
+	accounts.NewStateWriter(v, accounts.Contract.StateSubrealm(d)).SetInitialState(0, baseTokenCoinInfo)
```
nnikolash: NOTE:
Initial state of common account is now always 0, no matter what was the actual balance.

packages/solo/solo.go
```diff
@@ -242,7 +242,7 @@ func (env *Solo) GetChainByName(name string) *Chain {
 }
 
 const (
-	DefaultCommonAccountBaseTokens   = 50 * isc.Million
+	DefaultGasCoinBalance            = 50 * isc.Million
```
nnikolash, line 245: NOTE:
Now common account is always 0 from the start, so no sense in defining its defaul value. However, the same value was also used for gas coin, and this is kept (but renamed).

packages/vm/core/evm/evmtest/evm_test.go
```diff
@@ -2116,7 +2116,7 @@ func TestEVMGasPriceMismatch(t *testing.T) {
 	} {
 		t.Run(v.name, func(t *testing.T) {
 			env := InitEVM(t)
-			env.Chain.DepositBaseTokensToL2(10*isc.Million, nil)
+			env.Chain.DepositBaseTokensToL2(50*isc.Million, nil)
```
nnikolash, line 2119: NOTE:
I've increased that value to fix the test, because before that was 50+10 = 60 (`DefaultCommonAccountBaseTokens` + manual deposit of 10 in test).

packages/isc/consts.go
```diff
@@ -9,4 +9,8 @@ const (
 
 // TODO Add the comprehensive top up calculation logic, then we can remvoe this constant
 
-const TopUpFeeMin = iotaclient.DefaultGasBudget * 5
+// This threshold defines the amount of funds, which must always be enough to cover
+// the gas costs of chain state transition for any number requests within supported range.
+const maxTotalGasCostForChainStateTransition = iotaclient.DefaultGasBudget * 5
+
+const GasCoinMinBalance = maxTotalGasCostForChainStateTransition
```
nnikolash: NOTE:
Renamed it, because that value is not a "top up fee min". Top-up is calculated as the difference between min balance of the gas coin and its actual balance. So this value is min balance of gas coin.

packages/isc/consts.go
```diff
@@ -9,4 +9,8 @@ const (
 
 // TODO Add the comprehensive top up calculation logic, then we can remvoe this constant
 
-const TopUpFeeMin = iotaclient.DefaultGasBudget * 5
+// This threshold defines the amount of funds, which must always be enough to cover
+// the gas costs of chain state transition for any number requests within supported range.
+const maxTotalGasCostForChainStateTransition = iotaclient.DefaultGasBudget * 5
```
howjmay, line 12: curious why we need this value. For sure it is a good thing to have it, but it seemingly havent been used in other places either?

packages/vm/core/testcore/base_test.go
```diff
@@ -26,26 +26,23 @@ func TestInitLoad(t *testing.T) {
...
-		originAmount,
-		cassets.BaseTokens())
-	require.EqualValues(t, 1, len(cassets.Coins))
-
 	t.Logf("common base tokens: %d", ch.L2CommonAccountBaseTokens())
-	require.True(t, cassets.BaseTokens() >= governance.DefaultMinBaseTokensOnCommonAccount)
+
+	cassets := ch.L2CommonAccountAssets()
+	require.EqualValues(t, 0, cassets.BaseTokens())
+	require.EqualValues(t, 0, len(cassets.Coins))
```
nnikolash, line 37: NOTE:
Changed it, because now  common account has no tokens from the start.

packages/isc/consts.go
nnikolash: This value is used as explanation of how `GasCoinMinBalance` is formed:
```
// This threshold defines the amount of funds, which must always be enough to cover
// the gas costs of chain state transition for any number requests within supported range.
const maxTotalGasCostForChainStateTransition = iotaclient.DefaultGasBudget * 5

const GasCoinMinBalance = maxTotalGasCostForChainStateTransition
```
Maybe in this case it is truly useless... But I wanted to avoid same situation we have now with `DefaultMinBaseTokensOnCommonAccount `. We see it exists, we see how it works, but we don't know why we need it 😄 
Because right now it is like this:
```
	// DefaultMinBaseTokensOnCommonAccount can't harvest the minimum
	DefaultMinBaseTokensOnCommonAccount = 3000
```
But image if that was done like:
```
const StorageDeposit  = 123
const MaxTotalTransactionGasCost = 456
const DefaultMinBaseTokensOnCommonAccount = StorageDeposit + MaxTotalStateTransitionGasCost
```

packages/isc/consts.go
```diff
@@ -9,4 +9,8 @@ const (
 
 // TODO Add the comprehensive top up calculation logic, then we can remvoe this constant
```
dessaya: Remove or update this comment?

packages/origin/origin.go
dessaya: Maybe remove the parameter if it is always 0?

dessaya (approved):

packages/isc/consts.go
nnikolash: Not sure what this comment is about - what logic would be "comprehensive". For now I'll leave it as a bookmark for later review of TODOs.

PR #454, opened 2025-01-14
fix: Add ConsensusGasPriceProposal() to interface

howjmay:


packages/chain/node_test.go
```diff
@@ -385,6 +386,13 @@ func (tnc *testNodeConn) WaitUntilInitiallySynced(ctx context.Context) error {
 	panic("should be unused in test")
 }
 
+func (tnc *testNodeConn) ConsensusGasPriceProposal(
+	ctx context.Context,
+	anchor *isc.StateAnchor,
+) <-chan cons_gr.NodeConnGasInfo {
+	panic("TODO")
```
kape1395: Can you return some default here, please? :)

kape1395 (changes requested):
Would be nice to have a default value instead of a panic in the test. I will get stuck on that place immediately :)

PR #457, opened 2025-01-16
Update docker to use the newest iota-tools+potential ARM image

lmoe:


packages/testutil/l1starter/local_node.go
```diff
@@ -41,9 +42,14 @@ func NewLocalIotaNode(iscPackageOwner iotasigner.Signer) *LocalIotaNode {
 func (in *LocalIotaNode) start(ctx context.Context) {
 	in.ctx = ctx
 
+	image := "linux/amd64"
```
nnikolash: I think its more of a "platform" instead of "image".

nnikolash (approved):

PR #458, opened 2025-01-16
Fix lint issues and context issues

lmoe:
# Description of change

Fixes all linter issues for wasp-cli


tools/wasp-cli/chain/metadata.go
```diff
@@ -155,8 +156,8 @@ func updateMetadata(node string, chainAliasName string, chainID isc.ChainID, wit
 
 	if useCliURL {
 		apiURL := config.WaspAPIURL(node)
-		chainPath, err := url.JoinPath(apiURL, "/v1/chains/", chainID.String())
-		log.Check(err)
+		chainPath, err2 := url.JoinPath(apiURL, "/v1/chains/", chainID.String())
+		log.Check(err2)
```
nnikolash, line 160: Just thinking out loud:
If this change is from linter, maybe that linter should be disabled or nolint used, because using `err2` is even more dangerous then shadowing - you could just return `err2` but check `err`. Hm, but maybe linter will also catch that.

tools/wasp-cli/cli/keychain/keychain_file.go
```diff
@@ -13,7 +13,10 @@ import (
 	"github.com/iotaledger/wasp/packages/cryptolib"
 )
 
-const fileName = "secrets.db"
+const (
+	fileName                   = "secrets.db"
+	filePermission os.FileMode = 0o777
```
nnikolash, line 18: Maybe it maks sense to do some more informational name or add comment

tools/wasp-cli/util/types.go
```diff
@@ -87,7 +88,7 @@ func ValueFromString(vtype, s string, chainID isc.ChainID) []byte {
 			log.Fatal("invalid nftid length")
 		}
 
-		var nid = [iotago.AddressLen]byte(nidBytes)
+		nid := [iotago.AddressLen]byte(nidBytes)
```
nnikolash, line 91: Maybe it makes sense to disable this lint also - for me it seems as too much of moderation.

nnikolash:
How did you come up to the idea of fixing contexts?
I see, that the actual context is still always `Background()`, so I assume there was some other purpose. E.g., satisfy linter or prepare code for external developers?

nnikolash (approved):

lmoe:
We should add proper timeouts to our CLI which is impossible if all the functions we call create their own context. I agree with the linter there, to pass a context from the start to all functions. For the rest, yeah that's debateable.. :D

PR #463, opened 2025-01-17
test: Fix TestCounter

howjmay:


lmoe (approved):

packages/vm/core/testcore/sbtests/concurrency_test.go
```diff
@@ -21,6 +21,10 @@ func TestCounter(t *testing.T) {
 
 	req := solo.NewCallParamsEx(ScName, sbtestsc.FuncIncCounter.Name).AddBaseTokens(1 * isc.Million).WithGasBudget(math.MaxUint64)
 	for i := 0; i < 33; i++ {
+		// charge more fund to gas coin every 5 request, otherwise there will be no fund to post request on L1
+		if i%5 == 4 {
```
dessaya, line 25: Why not instead do something like:

```
if the gas coin has less than X tokens, then:
  top up the gas coin
```

lmoe: sorry for merging already :>

dessaya: Ah no problem, anyway I see that this change affects a single test, so it's not important

PR #464, opened 2025-01-17
chore: Removed error return from some BCS functions, Added more tests

nnikolash:


.golangci.yml
```diff
@@ -101,12 +106,10 @@ linters-settings:
         locale: US
     nolintlint:
         allow-unused: false # report any unused nolint directives
-    goimports:
```
lmoe, line 104: Why did you remove goimports here? It sorts the imports, which I think is good.

packages/util/bcs/uint128_test.go
```diff
@@ -35,3 +35,6 @@ func testUint128Codec(t *testing.T, v string, expectSuccess bool) {
 		require.Error(t, err)
 	}
 }
+
+func TestInt64asUint128(t *testing.T) {
```
lmoe: Can this be removed?

lmoe (approved):

dessaya (approved):

.golangci.yml
nnikolash: The problem it does not sort the imports - the sorter here is an irritated engineer 😁 I would love to have this, but not before we have formatter which does this automatically. What do you think?

nnikolash: I'm merging the code to avoid later conflicts, but nevertheless I'll be ready revert my change to goimports if we decide that we need it

PR #468, opened 2025-01-20
chore: scriptsizefix4 test branch

lmoe:


packages/webapi/controllers/corecontracts/accounts.go
```diff
@@ -23,10 +23,10 @@ func (c *Controller) getTotalAssets(e echo.Context) error {
 		return c.handleViewCallError(err)
 	}
 
+	nt := assets.NativeTokens()
 	assetsResponse := &models.AssetsResponse{
-		BaseTokens: assets.BaseTokens().String(),
-		// TODO: fix this when native tokens reimplemented
-		// NativeTokens: isc.NativeTokensToJSONObject(assets.NativeTokens),
+		BaseTokens:   assets.BaseTokens().String(),
+		NativeTokens: nt.JSON(),
```
dessaya, line 29: Shouldn't we remove the concept of native tokens? We removed it in isc package and it simplified a lot of code. Now we just have a collection of coins. One of those coins is the base token, but it is not a special case.

lmoe: @dessaya I wasn't so sure about it. Yes we now only have a map of coins. But in the API context, I found it useful to have a separation between the Base and the custom coins. So you don't have to sort them out by yourself.

But I have no hard feelings on either approach.

PR #471, opened 2025-01-21
Implement deducting the topup fee from the common account

lmoe:
Now there is a single parameter `GasCoinTargetValue` (configurable in governance).

Upon chain deploy, we create the L1 gas coin with `GasCoinTargetValue`, and the L2 common account may be left empty.

After the VM processes a request, the collected fees are used to top up the common account up to `GasCoinTargetValue`. The rest goes to the payout AgentID.

Before creating the PTB, the VM tops up the L1 gas coin up to `GasCoinTargetValue`, deducting as needed from the L2 common account.


packages/solo/solo_test.go
```diff
@@ -16,9 +16,10 @@ func TestMain(m *testing.M) {
...
 	env := solo.New(t, &solo.InitOptions{Debug: true, PrintStackTrace: true})
 	ch := env.NewChain(false)
-	require.EqualValues(env.T, solo.DefaultCommonAccountBaseTokens, ch.L2CommonAccountAssets().Coins.BaseTokens())
+	require.Zero(env.T, ch.L2CommonAccountAssets().Coins.BaseTokens())
+	require.Zero(env.T, ch.L2BaseTokens(ch.OriginatorAgentID))
 
 	err := ch.DepositBaseTokensToL2(solo.DefaultChainOriginatorBaseTokens, nil)
 	require.NoError(env.T, err)
-	require.EqualValues(env.T, solo.DefaultChainOriginatorBaseTokens, ch.L2BaseTokens(ch.OriginatorAgentID))
+	require.NotZero(env.T, ch.L2BaseTokens(ch.OriginatorAgentID))
```
nnikolash, line 24: Why this change? Isnt `NotZero` check too generic?

dessaya: The resulting L2 balance is whatever you deposit minus the gas fee. But there is already another test for that; I didn't want to repeat the same assertion.

packages/vm/core/testcore/base_test.go
```diff
@@ -60,18 +58,68 @@ func TestLedgerBaseConsistency(t *testing.T) {
...
+	require.EqualValues(
+		t,
+		vmRes.Receipt.GasFeeCharged-addedToCommonAccount,
+		ch.L2BaseTokens(ch.OriginatorAgentID),
+	)
+	require.EqualValues(
+		t,
+		isc.GasCoinTargetValue*4-vmRes.Receipt.GasFeeCharged,
+		ch.L2BaseTokens(isc.NewAddressAgentID(someUserAddr)),
+	)
```
nnikolash: Could you pls put 1-line comment to each of these three requrements? That would help validate the logic

packages/vm/core/testcore/base_test.go
```diff
@@ -60,18 +58,68 @@ func TestLedgerBaseConsistency(t *testing.T) {
...
+		isc.GasCoinTargetValue-coin.Value(ptbRes.Effects.Data.GasFee()),
+		gasCoinValueAfter,
+	)
+
+	totalDeposited := coin.Value(isc.GasCoinTargetValue/2 + isc.GasCoinTargetValue*4)
+	totalDeductedForGasCoin := isc.GasCoinTargetValue - gasCoinValueBefore
+	totalL2Tokens := totalDeposited - totalDeductedForGasCoin
+	require.EqualValues(t, coin.Value(totalL2Tokens), ch.L2TotalBaseTokens())
+
+	addedToCommonAccount := min(isc.GasCoinTargetValue, vmRes.Receipt.GasFeeCharged)
```
nnikolash: This line feels strange. Could you pls add comment why is it so?
Because for me it feels like it should be like:
```suggestion
	addedToCommonAccount := min(isc.GasCoinTargetValue - gasCoinValueBefore, vmRes.Receipt.GasFeeCharged)
```
Like, why would we deduct from fee and put on common account the entire `isc.GasCoinTargetValue` if it was not fully used?
Damn, its hard to think about this stuff in this late time

packages/vm/vmimpl/runtask.go
```diff
@@ -43,23 +42,33 @@ func newVmContext(
...
+	if bal < topUp {
+		vmctx.task.Log.Warnf(
+			"not enough tokens in common account for topping up gas coin (has %d, want %d)",
+			bal,
+			topUp,
+		)
 	}
 
-	return requiredTopUp
+	topUp = min(bal, topUp)
```
nnikolash: this cut be put under `if bal < topUp {`

nnikolash (dismissed):
The test is work of art 🔥

packages/vm/core/testcore/base_test.go
```diff
@@ -60,18 +58,87 @@ func TestLedgerBaseConsistency(t *testing.T) {
...
+	// the user pays for the L2 gas fee
+	require.EqualValues(t,
+		depositedByUser-vmRes.Receipt.GasFeeCharged,
+		ch.L2BaseTokens(isc.NewAddressAgentID(someUserAddr)),
+	)
+
+	// the common account is topped up to GasCoinTargetValue, unless the
+	// collected fees are less than that
+	addedToCommonAccount := min(
+		isc.GasCoinTargetValue-initialCommonAccountBalance,
```
nnikolash, line 127: Are you sure it's now correct? Why we top up gas coin for the difference between its desired balance and balance of common account? Shouldn't it be difference between gas coin actual balance and desired balance as in my example from previous comment?

nnikolash: Oh, do we use balance of common account as an indication of gas coin actual balance? Is that the reason?

dessaya: Yes. We now have two steps:
- first, collect all L2 fees into common account, up to GasCoinTargetValue
- second, transfer from common account to gas coin

In the first step, the balance of the gas coin is not relevant.

PR #482, opened 2025-01-23
refactor: Refactor metrics

howjmay:


lmoe (commented):
Good choices.

I have seen you commented out InAnchor, I would include it.
I would also include Milestone/Checkpoint
Remember that TransactionID is not an ID but a Digest now

The rest seems sensible. Please finish on that. :)

howjmay:
I thought checkpoint we can easily get it by transaction digest and InAnchor I move the related data to `outPublishStateTransaction` which store isc.StateAnchor now. Not sure if this way works

lmoe (approved):

lmoe:
LGTM If you tested it properly and consider it functioning, feel free to merge.

PR #483, opened 2025-01-23
Update move contract

lmoe:
Updates move contracts to 0.9.0
Fixes Anchor decoding issue


clients/iscmove/iscmoveclient/iscmovetypes.go
```diff
@@ -32,7 +27,7 @@ type MoveRequest struct {
 	ID     iotago.ObjectID
 	Sender *cryptolib.Address
 	// XXX balances are empty if we don't fetch the dynamic fields
-	AssetsBag Referent[iscmove.AssetsBagWithBalances] // Need to decide if we want to use this Referent wrapper as well. Could probably be of *AssetsBag with `bcs:"optional`
+	AssetsBag iscmove.Referent[iscmove.AssetsBagWithBalances] // Need to decide if we want to use this Referent wrapper as well. Could probably be of *AssetsBag with `bcs:"optional`
```
howjmay: We can remove the comment here too btw, since it is Move types, so Referent must be there

howjmay (approved):

howjmay:
Not sure whether the error is caused by this PR

lmoe:
it is :D
it happend right before i had to jump in a call, will finish it soon

PR #488, opened 2025-01-27
Make ChainMgr use a bit more event-based instead of polling.

kape1395:
Should reduce logging and also the CPU usage for re-checking the cmtlog output.


lmoe (approved):

kape1395:
@lmoe, have I broken the tests, or is it still the norm to have them fail?

howjmay:
Looks like other error
```
testdbhash.go:111: State hash for core contract "governance" has changed! This may be due to a BREAKING CHANGE; make sure that you add a migration (if necessary), and then run all tests again with: UPDATE_DBHASHES=1 (e.g. `UPDATE_DBHASHES=1 make test`). Note: you can set DB_DUMP=1 in one branch and DB_DUMP=2 on another, and then compute a diff of the generated hex dumps.
```

lmoe:
Hmm. I don't remember seeing tests fail there. I'm unsure if the wrong blockindex error is intended in the tests. (As you poll to wait for a new block, or so)

But it seems something is broken I havent seen before. It doesnt seem like its caused by your PR except for the message above with the different state hash that Yang Hao pointed out (which is fine, but must be adjusted with the env variable in your branch)

dessaya:
For now you can ignore the warning about the DB hash mismatch. It is not an error, just a warning.

PR #489, opened 2025-01-28
Fix cli

vitaliy-io:


tools/wasp-cli/util/tx.go
```diff
@@ -34,6 +35,10 @@ func WithSCTransaction(ctx context.Context, client *apiclient.APIClient, chainID
 	log.Check(err)
 	log.Printf("Posted on-ledger transaction %s\n", tx.Digest)
 
+	ref, err := tx.GetCreatedObjectInfo(iscmove.RequestModuleName, iscmove.RequestObjectName)
+	log.Check(err)
+	log.Printf("Request ID: %s\n", isc.RequestID(ref.ObjectID.Data()).String())
```
howjmay: maybe using?
```
log.Printf("Request ID: %s\n", ref.ObjectID.ShortString())
```

vitaliy-io: I see that ShortString trims zeroes on the left, but why would we need to trim zeroes in fixed size address?
I guess we could just use ref.ObjectID.String() instead?

howjmay: Either way is ok. I was thinking, a shorter strong may be more readable

vitaliy-io: fixed

lmoe (approved):

PR #496, opened 2025-02-02
fix: Check iotago.GetObject error

howjmay:


clients/iota-go/iotaclient/api_read.go
```diff
@@ -47,7 +47,13 @@ type GetObjectRequest struct {
 
 func (c *Client) GetObject(ctx context.Context, req GetObjectRequest) (*iotajsonrpc.IotaObjectResponse, error) {
 	var resp iotajsonrpc.IotaObjectResponse
-	return &resp, c.transport.Call(ctx, &resp, getObject, req.ObjectID, req.Options)
+	err := c.transport.Call(ctx, &resp, getObject, req.ObjectID, req.Options)
```
dessaya, line 51: Shouldn't the same fix be applied to all functions that return `IotaObjectResponse`? For example `GetDynamicFieldObject`, `MultiGetObjects`, etc.

howjmay: all updated

dessaya (approved):

PR #499, opened 2025-02-04
Add ws reconnection

vitaliy-io:


clients/iota-go/iotaconn/websocket.go
```diff
@@ -47,6 +50,8 @@ func (c *WebsocketClient) WaitUntilStopped() {
 	c.shutdownWaitGroup.Wait()
 }
 
+// var counter = 0
```
howjmay: dead var

clients/iota-go/iotaconn/websocket.go
```diff
@@ -229,3 +318,111 @@ func (c *WebsocketClient) nextID() string {
...
 	return strconv.FormatUint(uint64(id), 10)
 }
+
+func (c *WebsocketClient) reconnect(ctx context.Context) error {
+	c.log.Debugf("reconnecting")
+	if c.reconnectMx.TryLock() {
+		defer c.reconnectMx.Unlock()
+	} else {
+		// already reconnecting, try again later
+		time.Sleep(50 * time.Millisecond)
```
nnikolash, line 321: If one thread hangs on reconnect, another will create 1 stack frame every 50ms, because e.g. `writeMessage` uses recursion after `reconnect` returns. Can this lead to stack overflow eventually?

clients/iota-go/iotaconn/websocket.go
```diff
@@ -8,36 +8,39 @@ import (
...
 
 type WebsocketClient struct {
 	idCounter         uint32
+	url               string
 	conn              *websocket.Conn
 	writeQueue        chan *jsonrpcMessage
 	readers           sync.Map // id -> chan *jsonrpcMessage
 	log               *logger.Logger
 	shutdownWaitGroup sync.WaitGroup
+	reconnectMx       sync.RWMutex
```
nnikolash: You use RWMutext but I do not see any `RLock`/`RUnlock`

clients/iota-go/iotaconn/websocket.go
```diff
@@ -8,36 +8,39 @@ import (
...
-	}
+
 	c := &WebsocketClient{
-		conn:       conn,
+		url:        url,
 		writeQueue: make(chan *jsonrpcMessage),
 		log:        log,
 	}
+
+	c.reconnect(ctx)
```
nnikolash: Minor thing, not a required fix: Maybe check an error? If context is canceled, if would be nice to return error, especially when we already have error returned from the contructor.

clients/iota-go/iotaconn/websocket.go
```diff
@@ -229,3 +318,111 @@ func (c *WebsocketClient) nextID() string {
...
+
+		method := sub.method
+		args := sub.args
+		oldId := sub.id
+
+		var subID uint64
+		err := c.CallContext(ctx, &subID, method, args...)
+		if err != nil {
+			c.log.Errorf("failed to resubscribe to %s: %s", method, err)
+			continue
```
nnikolash, line 405: Maybe we should return here with error? Why would we support subscribing for part of all subscriptions? Wouldn't hta be a faulty state?

clients/iota-go/iotaconn/websocket.go
```diff
@@ -229,3 +318,111 @@ func (c *WebsocketClient) nextID() string {
...
+		}
+		newId := fmt.Sprintf("%s:%d", method, subID)
+
+		// store reader channel with new id
+		ch, ok := c.readers.Load(oldId)
+		defer c.readers.Delete(oldId)
```
nnikolash: Why defer is here?

clients/iota-go/iotaconn/websocket.go
```diff
@@ -140,6 +194,23 @@ func (c *WebsocketClient) writeMsg(method JsonRPCMethod, args ...interface{}) (s
...
+
+var subscriptions = make([]*subscription, 0, 2)
+
+type call struct {
+	method JsonRPCMethod
+	args   []interface{}
+	id     string
+}
+
+var pendingCalls sync.Map
```
nnikolash: TLDR:
Shouldn't each client have its work `pendingCalls` and `subscriptions`?

--

Sorry, might be a stupid question, because I just did not fully percieved the implementaion.
I see that `pendingCalls` is a global variable. Could you please tell couple words about the reason?
Now seems that it will be shared between multiple clients. And these clients may have their own `url` and their own pending messages. So how that would work in following case: Client1 adds some pending calls. Client2 adds more pending calls, then for Client3 connection fails and it starts to reconnect and when succeeds it sends all the pending messages from all the clients?

clients/iota-go/iotaconn/websocket.go
```diff
@@ -229,3 +318,111 @@ func (c *WebsocketClient) nextID() string {
...
+	if c.reconnectMx.TryLock() {
+		defer c.reconnectMx.Unlock()
+	} else {
+		// already reconnecting, try again later
+		time.Sleep(50 * time.Millisecond)
+		return nil
+	}
+
+	if c.conn != nil {
+		c.conn.Close()
```
nnikolash, line 326: If read and write both will need to reconnect, would that mean that eventually they both will call this function?
I mean:
1) read fail and tried to reconnect
2) write fails and tries to reconnect but is required to wait for read to reconnect
3) read reconnects -> resubscribes and recreated pending calls
4) write "unfreezes" and is now also able to reconnect, and thus deletes again the connect (which might again trigger reconnect for read), and again resubscribes and recreates calls

I might be missing something

vitaliy-io:
will address all questions/suggestions if we actually end up using this solution tomorrow

clients/iota-go/iotaconn/websocket.go
vitaliy-io: I initially thought that this client is a singleton but I see now that it's per chain, so yes, needs to be per client

vitaliy-io: actually no, it's not per chain, but a singleton :). But anyway, making this map per client seems cleaner, so fixed that

clients/iota-go/iotaconn/websocket.go
vitaliy-io: agree. fixed

clients/iota-go/iotaconn/websocket.go
vitaliy-io: not sure. I though it makes sense to resubscribe to what we can anyway. Worst case scenario is that it just does not work

clients/iota-go/iotaconn/websocket.go
vitaliy-io: fixed

clients/iota-go/iotaconn/websocket.go
vitaliy-io: It was needed before. Now it's not. Fixed

clients/iota-go/iotaconn/websocket.go
vitaliy-io: I guess in theory we could run out of memory but it will only happen if L1 is not accessible for a really long time, which means that wasp will not be working anyway. Besides, golang's stack is more flexible and can grow dynamically. So I'm not sure if it's a major issue

clients/iota-go/iotaconn/websocket.go
vitaliy-io: I think this scenario is possible but is very unlikely, it could only happen if second flow (write in your case) calls reconnect RIGHT after the first reconnect finishes and the lock is released. In most cases second connection will just be waiting 50ms and then reconnect using new connection

clients/iota-go/iotaconn/websocket.go
vitaliy-io: fixed

PR #505, opened 2025-02-11
Rotation can be initiated by a quorum of an existing committee nodes.

kape1395:


packages/chain/cons/cons.go
```diff
@@ -519,6 +534,19 @@ func (c *consImpl) uponACSOutputReceived(outputValues map[gpa.NodeID][]byte) gpa
 	baoID := bao.GetObjectID()
 	reqs := aggr.DecidedRequestRefs()
 	c.log.Debugf("ACS decision: baseAO=%v, requests=%v", bao, reqs)
+	if aggr.DecidedRotateTo() != nil {
+		c.log.Debugf("Will rotate to %v", aggr.DecidedRotateTo().ToHex())
+		rotationPTB := vmtxbuilder.NewAnchorTransactionBuilder(bao.ISCPackage(), bao, c.dkShare.GetAddress())
+		rotationPTB.RotationTransaction(aggr.DecidedRotateTo())
+		rotationPTX := rotationPTB.BuildTransactionEssence(bao.GetStateMetadata(), 0)
+		rotationTXD := c.makeTransactionData(&rotationPTX, aggr)
+		rotationTXB := c.makeTransactionBytes(rotationTXD)
```
dessaya, line 542: So the rotation is no longer responsibility of the VM? In this case we can remove al code related to rotation from there?

dessaya (approved):

packages/chain/cons/cons.go
kape1395: I guess so.

lmoe (approved):
Good stuff!

PR #506, opened 2025-02-11
feat: Transfer GasCoin ownership in the end of rotation

howjmay:


dessaya (approved):

kape1395:
Shouldn't we do this logic in the MOVE contact and only call it here as a function?

howjmay:
@kape1395 we may meet some problem when transferring GasCoin in move contract. 
In the current txbuilder, we transfer the GasCoin in the end of all PTB. Therefore, we transfer Anchor's ownership then transfer GasCoin's ownership after. 
However, if we integrate it into Move contract, then we will transfer GasCoin's ownership before we transfer Anchor's ownership. 
I can try to do some experiment to check this behavior

PR #507, opened 2025-02-13
fix: Use Digest to stand for TXs instead of HashValue

howjmay:


clients/iota-go/iotago/byte_data_types_test.go
```diff
@@ -18,8 +19,11 @@ func TestSerialization(t *testing.T) {
 
 	base64data := iotago.Bytes(hexdata.Data()).GetBase64Data()
 	base64Str := base64data.String()
+	b64Bcs, err := bcs.Marshal(&base64data)
+	require.NoError(t, err)
 
 	t.Log(base64Str)
+	t.Log(b64Bcs)
 	t.Log(hexStr)
```
dessaya: why not require.Equal instead of logging?

clients/iota-go/iotago/digest.go
```diff
@@ -12,22 +22,72 @@ type (
...
+}
+
+func (d Digest) MarshalBCS(e *bcs.Encoder) error {
+	e.Encode(d[:])
+	return nil
+}
+
+func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
+	b := bcs.Decode[[33]byte](de)
+	copy(d[:], b[1:])
```
dessaya: Pls avoid magic numbers. Why is the extra byte needed?

howjmay: According to BCS spec, fixed-length array doesn't have a leading element to indicate the size. 
All the Digest are in 32 bytes. However, they are treated as variable-length sequences in MoveVM instead of a fixed-length array. This means in MoveVM, there is always a leading byte which is 32 for the BCS encoding of Digests.
Here I only take the part of the slice I need which is the last 32 bytes, and get rid of the leading byte.

I have added comment for it

clients/iota-go/iotago/byte_data_types_test.go
howjmay: fixed

clients/iota-go/iotago/digest.go
```diff
@@ -12,22 +22,75 @@ type (
...
+	e.Encode(d[:])
+	return nil
+}
+
+func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
+	b := bcs.Decode[[DigestSize + 1]byte](de)
+	// Digests are stored as variable-length sequences in MoveVM
+	// To adpat to a fixed-legnth array (32 bytes), we need to get rid of the leading element from BCS
+	// The leading element is always 32, which represents the legnth of the variable-length sequences
+	copy(d[:], b[1:])
```
dessaya: Why is this special logic needed for `UnmarshalBCS` but not for `MarchalBCS`?

If the digest is encoded as `[]byte` (instead of `[DigestSize]byte`) then why not decode it as `[]byte`? Like `bcs.Decode[[]byte](de)`?

howjmay: `Digest`s are stored as variable-length sequences in MoveVM, so the BCS encoding for TXs (or any other objects interact with nodes) should have the leading byte to describe the length of the sequences. Therefore, the encoding is 32+1 bytes.
However, in reality, `Digest`s are always 32 bytes, so in Go implementation if we declare `Digest` as the alias of [32]byte, we need to exclude the leading byte to adopt the difference between MoveVM and Go implementation

dessaya: What I mean is that `MarshalBCS` is implemented as `e.Encode(d[:])` which encodes as `[]byte`.

So, for consistency, we should also decode as `[]byte`:

```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
  b := bcs.Decode[[]byte](de)
  if len(b) != len(d) {
    return errors.New("unexpected length")
  }
  copy(d[:], b)
  return nil
}
```

howjmay: How about
```go
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
  b := bcs.Decode[[]byte](de)
  if len(b) != len(d)+1 {
    return errors.New("unexpected length")
  }
 copy(d[:], b[1:])
  return nil
}

```

dessaya: I don't think that would work. BCS should automatically strip the extra byte when decoding `[]byte`.

Or maybe i'm not understanding the problem. @nnikolash maybe can help?

howjmay: The encoding of fixed length array is different from variable-length sequences.
The encoding of a `[32]byte` is 32 byte, but the encoding of a `[]byte` slice with 32 byte has 33 byte.
So I have to exclude the leading byte to adopt to the fixed length array in Go-SDK.

In MoveVM: use variable-length sequences for Digest (33 bytes in BCS)
In Go-SDK: use fixed length array sequences for Digest  (32 bytes in BCS)

nnikolash: I think Yanhao's understanding of the problem is correct - if in the code we have fixed-array, but for some reason length must be also encoded, custom serialization functions is way to go.
Regarding the implementation, I think both Yangao's and Diego's implementations are correct. You can choose any of your taste.

BUT that is an interesting case. Because after seeing this, I thought how can we avoid copying bytes here? And to my dissapointment it seems, that we can't - not because there is some conceptual issue, but just because the lib is not implemented for that.
Here is why:
1. Decoding into slice taked from array:
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
	de.Decode(d[:])
	return nil
}
```
The issue here is that library is specifically designed to not "reuse" slices - it always creates new empty slice and fills it. This is done to avoid unexpected bugs. So in this case nothing will be changed at all, because slice here is just a temporary non-addressable variable.

2. Decoding into pointer to a slice taked from array: 
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
	de.Decode(&d[:])
	return nil
}
```
The issue here is that we pass address of temporary slice variable into `Decode`. Setting that variable does not change anything.

3. Skipping length and then readin into an array:
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
	_ = de.ReadLen()
	de.Decode(d)
	return nil
}
```
Here we get recursion, because `d` is `Digest`, and decoding it will invoke `UnmarshalBCS`, in which we already are.

All three cases can be implemented if somebody will invest time into it.

nnikolash: Oh, btw for this particular case (`[]byte`) there IS a way to avoid copying bytes:
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
	_ = de.ReadLen()
	_, _ = de.Read(d[:])
	return nil
}
```

dessaya: Thanks @nnikolash !

I still don't understand why there is an asymmetry between the logic needed in `MarchalBCS` and `UnmarshalBCS`.

In my mind, we have a `[32]byte` but we want it to encode as `sizeByte + [32]byte`. So in `MarshalBCS` we encode it as `[]byte` which does exactly that. And for consistency in `UnmarshalBCS` we can just decode a `[]byte`.

The extra step in `UnmarshalBCS` looks suspicious. Why is it needed?

nnikolash: @dessaya When I was saying, that your implementation and Yanghao's are both valid, I meant his implementaion from the code, and not one from comments. The one from comments is incorrect I think, because that +1 won't work.

So I was comparing these two:
* From the code:
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
	b := bcs.Decode[[DigestSize + 1]byte](de)
	copy(d[:], b[1:])
    return nil
```
* Yours from comments:
```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
    b := bcs.Decode[[]byte](de)
    copy(d[:], b)
    return nil
}
```
They do smae thing, just different perspective:
* Yanghao skipping length by reading it as a first element of array. Which is not very clear, because what if array is longer then 255? But here it is not a problem and would even be more performant implementation :)
* Your implementation skips length by reading it as slice length.

Overall, effect is the same.

howjmay: @dessaya what is your decision for this change? If you think it is better to change to decode into slice then I will do it here

dessaya: IMO it is better to favor code clarity. So:

```
func (d *Digest) UnmarshalBCS(de *bcs.Decoder) error {
    b := bcs.Decode[[]byte](de)
    copy(d[:], b)
    return nil
}
```

dessaya (approved):

PR #508, opened 2025-02-17
Use http instead of ws for calls

vitaliy-io:


howjmay:
Isnt this fetch all the events in the past? Maybe it will be enormous amount of events will be returned? 
I think we need to talk with devop team. It looks like TimeRange filter is not enabled. Or you can give it a try

lmoe:
> Isnt this fetch all the events in the past? Maybe it will be enormous amount of events will be returned? I think we need to talk with devop team. It looks like TimeRange filter is not enabled. Or you can give it a try

Care to elaborate on this?
If I understand you correctly you mean Requests, not Events? Right now the implementation is expected to load all pending request IDs, to sort them and then select n requests. Right now 1000. I tested it up to 25k which worked fine. But if you mean something else, please tell. :D

lmoe (approved):

packages/nodeconn/chain.go
```diff
@@ -76,10 +84,10 @@ func newNCChain(
 
 	// FIXME make timeout configurable
 	// FIXME this will be replaced by passing l1param from consensus
-	l1syncer := parameters.NewL1Syncer(nodeConn.wsClient.Client, 600*time.Second, nodeConn.Logger())
+	l1syncer := parameters.NewL1Syncer(nodeConn.httpClient.Client, 600*time.Second, nodeConn.Logger())
```
howjmay: I think `parameters.NewL1Syncer()` is outdated

howjmay:
@lmoe You are right. I misunderstood the PR. Te usage of httpClient and wsClient are correct!

packages/nodeconn/chain.go
vitaliy-io: I don't really know. I just refactored the field name wherever it was present

vitaliy-io: oh I see you've made these changes in master. I merged it and overrode my changes

PR #509, opened 2025-02-19
fix: trie iteration not stopping when callback returns false

dessaya:


nnikolash (approved):
Heh, saw this also, but at that time I did not know it was because of trie - I thought this was just a rocksb thing

PR #510, opened 2025-02-21
test: Fix TestCreateAndSendRequest

howjmay:


clients/iota-go/iotaclient/extend_calls.go
```diff
@@ -222,38 +222,81 @@ func (c *Client) MintToken(
...
-	)
-	if err != nil {
-		return nil, fmt.Errorf("can't execute the transaction: %w", err)
+	return c.SignAndExecuteRetryableTx(ctx, signer, pt, nil, DefaultGasBudget, DefaultGasPrice, options)
+}
+
+// The assigned gasPayments, or the gasPayments got by FindCoinsForGasPayment may be outdated ObjectRef
+// which would cause the execution of tx failed.
+// This func can retry a few time
+func (c *Client) SignAndExecuteRetryableTx(
```
dessaya: `SignAndExecuteTxWithRetry`?

Also, instead of making a `WithRetry` variant for every function maybe it's better to make a generic `DoWithRetry(func() bool)` or similar.

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
```diff
@@ -28,6 +28,15 @@ func NewRandomSignerWithFunds(t *testing.T, index int) cryptolib.Signer {
 	return NewSignerWithFunds(t, seed[:], index)
 }
 
+var genSignerCounter = 0
```
dessaya: Using a global variable will make test execution depend on each other.

Better: specify the index explicitly (as before), or use something that is unique for each test, for example `t.Name()`.

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
```diff
@@ -28,6 +28,15 @@ func NewRandomSignerWithFunds(t *testing.T, index int) cryptolib.Signer {
...
+
+func GenSignerWithFundByCounter(t *testing.T) cryptolib.Signer {
```
dessaya: `Fund` -> `Funds`

Also, why `Gen` instead of `New`? It's better to use consistent naming.

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
howjmay: I will try to use `t.Name()`. This change was aimed to give each tests an unique signer

clients/iota-go/iotaclient/extend_calls.go
howjmay: I am thinking maybe add the loop in each function is easier than adding the generic. Not sure how to design the proper interface to return the different condition and return value

packages/chain/node_test.go
```diff
@@ -135,12 +135,6 @@ func testNodeBasic(t *testing.T, n, f int, reliable bool, timeout time.Duration,
 				Function: uint32(mmm.Target.EntryPoint),
 				Args:     mmm.Params,
 			},
-			// inccounter.FuncIncCounter.Message(nil),
-			// &iscmove.Message{
-			// 	Contract: uint32(inccounter.Contract.Hname()),
-			// 	Function: uint32(inccounter.FuncIncCounter.Hname()),
-			// 	Args:     inccounter.FuncIncCounter.Message(nil).Params.Clone(),
-			// },
```
dessaya, line 143: this left commented on purpose?

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
```diff
@@ -66,7 +68,7 @@ func newSoloTestEnv(t testing.TB) *soloTestEnv {
 		PrintStackTrace: true,
 		Log:             log,
 	})
-	chainOwner, _ := s.NewKeyPairWithFunds()
+	chainOwner, _ := iscmoveclienttest.NewRandomSignerByTestName(t.(*testing.T)).(*cryptolib.KeyPair)
```
dessaya: The cast will fail when executed from a benchmark (`*testing.B`).

clients/iscmove/iscmoveclient/iscmoveclienttest/setup.go
```diff
@@ -28,6 +29,11 @@ func NewRandomSignerWithFunds(t *testing.T, index int) cryptolib.Signer {
 	return NewSignerWithFunds(t, seed[:], index)
 }
 
+func NewRandomSignerByTestName(t *testing.T) cryptolib.Signer {
+	nameHash := blake2b.Sum256([]byte(t.Name()))
```
dessaya: The created signer is deterministic, so calling it `Random` is confusing.

packages/origin/origin_test.go
```diff
@@ -47,8 +47,8 @@ func TestOrigin(t *testing.T) {
 
 func TestCreateOrigin(t *testing.T) {
 	client := iscmoveclienttest.NewHTTPClient()
-	sentSigner := iscmoveclienttest.NewRandomSignerWithFunds(t, 0)
-	stateSigner := iscmoveclienttest.NewRandomSignerWithFunds(t, 1)
+	sentSigner := iscmoveclienttest.NewRandomSignerByTestName(t)
+	stateSigner := iscmoveclienttest.NewRandomSignerByTestName(t)
```
dessaya: If I understand correctly this will create two identical signers? The intention is to have two different signers.

packages/chain/node_test.go
howjmay: But it does the same thing as the parameters above. The same contract and endpoint. It is a little redundant to leave it here

dessaya: Ah sorry I read it too fast and I thought the code was not removed. It's good to remove it :+1:

dessaya (approved):

PR #515, opened 2025-02-25
Add Get Object retry

lmoe:


clients/iscmove/iscmoveclient/client_anchor.go
```diff
@@ -23,6 +25,32 @@ type StartNewChainRequest struct {
 	GasBudget         uint64
 }
 
+func (c *Client) TryGetObject(ctx context.Context, req iotaclient.GetObjectRequest) (*iotajsonrpc.IotaObjectResponse, error) {
```
dessaya: Nit: Maybe call it `GetObjectWithRetry` or something like that? `TryGetObject` does not imply that it retries on failure.

dessaya (approved):

PR #517, opened 2025-02-27
Fix/wasp cli config order

vitaliy-io:


tools/wasp-cli/main.go
```diff
@@ -37,6 +39,12 @@ func initRootCmd(waspVersion string) *cobra.Command {
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			config.Read()
 
+			client := iotaclient.NewHTTP(config.L1APIAddress(), iotaclient.WaitForEffectsDisabled)
+			err := parameters.InitL1(*client, log.HiveLogger())
+			if err != nil {
+				panic(err)
+			}
```
dessaya: Prefer `log.Check(err)` to print user-facing errors

dessaya (approved):

tools/wasp-cli/main.go
vitaliy-io: fixed

PR #538, opened 2025-03-12
fix: Deleting remainder value from store if it is 0

nnikolash:


packages/vm/core/accounts/basetokens.go
```diff
@@ -77,5 +77,9 @@ func (s *StateReader) getWeiRemainder(accountKey kv.Key) *big.Int {
 }
 
 func (s *StateWriter) setWeiRemainder(accountKey kv.Key, v *big.Int) {
-	s.state.Set(accountWeiRemainderKey(accountKey), codec.Encode(v))
+	if v.Cmp(new(big.Int)) == 0 {
```
dessaya: Nit: I like to do `v.Sign() == 0` in this case to avoid creating a new Int, but maybe it's overoptimizing

dessaya (approved):

packages/vm/core/accounts/basetokens.go
nnikolash: No no, nice point! I'll fix that

PR #542, opened 2025-03-13
Fix tests2

lmoe:
Tries to fix some reliability issues in the tests. For now a test PR mainly to test the Github runner execution.


packages/vm/core/testcore/accounts_test.go
```diff
@@ -481,9 +481,9 @@ func initDepositTest(t *testing.T, initCommonAccountBaseTokens ...coin.Value) *a
 	ret := &accountsDepositTest{}
 	ret.env = solo.New(t, &solo.InitOptions{Debug: true, PrintStackTrace: true})
 
-	ret.chainOwner, ret.chainOwnerAddr = ret.env.NewKeyPairWithFunds(ret.env.NewSeedFromIndex(10))
+	ret.chainOwner, ret.chainOwnerAddr = ret.env.NewKeyPairWithFunds(ret.env.NewSeedFromTestName(t.Name()))
 	ret.chainOwnerAgentID = isc.NewAddressAgentID(ret.chainOwnerAddr)
-	ret.user, ret.userAddr = ret.env.NewKeyPairWithFunds(ret.env.NewSeedFromIndex(11))
+	ret.user, ret.userAddr = ret.env.NewKeyPairWithFunds(ret.env.NewSeedFromTestName(t.Name()))
```
howjmay: I guess here `ret.chainOwner` and ret.user` would be the same? They are hash value from the same string

lmoe: No, it hashes the function name and a timestamp internally. It's quite opaque. I'll rename the function.

PR #551, opened 2025-03-18
feat: Impl onledger request estimation

howjmay:


howjmay:
Run `make apiclient-docker` but the change was too much out of my change in this PR, so I revert that part.

packages/webapi/models/requests.go
```diff
@@ -15,7 +15,8 @@ type ContractCallViewRequest struct {
 }
 
 type EstimateGasRequestOnledger struct {
-	Output string `json:"outputBytes" swagger:"desc(Serialized Output (Hex)),required"`
+	DryRunRes string `json:"dryRunRes" swagger:"desc(The result of the DryRun encoded in BCS format and represented in hexadecimal),required"`
```
howjmay: Asked the user to call DryRun, because not sure if it is ok to access L1Client in `packages/webapi/controllers/chain/estimategas.go`

clients/iota-go/iotago/language_storage.go
```diff
@@ -132,6 +132,15 @@ func parseStructTypeArgs(structTypeParams string) ([]TypeTag, error) {
 	return retTypeTag, nil
 }
 
+func (s *TypeTag) UnmarshalJSON(data []byte) error {
+	tag, err := TypeTagFromString(string(data)[1 : len(data)-1])
```
dessaya: This is not properly parsing json. E.g. the string might contain escape sequences. Use `json.Unmarshal` to unmarshal the string and then call `TypeTagFromString`.

clients/iota-go/iotaclient/iotaclienttest/api_transaction_builder_test.go
```diff
@@ -124,7 +124,9 @@ func TestMoveCall(t *testing.T) {
 		},
 	)
 	require.NoError(t, err)
-	queryEventsResMap := queryEventsRes.Data[0].ParsedJson.(map[string]interface{})
+	queryEventsResMap := make(map[string]interface{})
```
dessaya: Avoid allocating the map twice:
```suggestion
	var queryEventsResMap map[string]any
```

packages/isc/request_onledger.go
```diff
@@ -58,6 +65,150 @@ func OnLedgerFromRequest(request *iscmove.RefWithObject[iscmove.Request], anchor
 	return r, nil
 }
 
+func FakeEstimateOnLedger(dryRunRes *iotajsonrpc.DryRunTransactionBlockResponse, msg *iscmove.Message) (OnLedgerRequest, error) {
+	assets := NewAssets(0)
+	allowance := NewAssets(0)
+	tx := dryRunRes.Input.Data.V1.Transaction.Data.ProgrammableTransaction
+	cmds := gjson.ParseBytes(tx.Commands)
```
dessaya: why use this gjson instead of json.Unmarshal?

lmoe (approved):

packages/isc/request_onledger.go
howjmay: The content here is dynamic and hard to declare a static type for it. So I use gjson for parsing it

howjmay:
Cluster tests failed mostly. I am still trying to verify the change

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
-		OutputBytes: cryptolib.EncodeHex(outputBytes),
+	// Estimate L2 gas budget for that transaction
+	estimatedReceipt, _, err := env.Chain.Cluster.WaspClient(0).ChainsAPI.EstimateGasOnledger(context.Background()).Request(apiclient.EstimateGasRequestOnledger{
+		TransactionBytes: hexutil.Encode(txBytes),
 	}).Execute()
 	require.NoError(t, err)
 	require.Empty(t, estimatedReceipt.ErrorMessage)
 
-	keyPair, _, err := env.Clu.NewKeyPairWithFunds()
+	l2GasBudget, err = strconv.ParseUint(estimatedReceipt.GasFeeCharged, 10, 64)
```
chriscandless: I'm confused regardin `l2GasBudget`. I thought `EstimateGasOnledger` should return total gas both for L1 and L2, not just L2?
Because otherwise the client should run dry run for himself to find L1 part, and then we also run dry run to get effects for decoding?

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
-	require.NoError(t, err)
+	// Create and execute same transaction, but with proper L2 gas budget
+	ptb = iscmoveclient.PTBCreateAndSendRequest(
+		assetsBagPTB.Clone(),
+		l1starter.ISCPackageID(),
+		env.Chain.ChainID.AsObjectID(),
+		argAssetsBag,
+		msg,
+		bcs.MustMarshal(allowance),
+		l2GasBudget,
```
chriscandless: What does this argument? If I pass 0 it still works with no problem

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
+	)
+	pt := ptb.Finish()
+
+	txData := iotago.NewProgrammable(
+		sender.Address().AsIotaAddress(),
+		pt,
+		[]*iotago.ObjectRef{},
+		2*iotaclient.DefaultGasBudget,
+		iotaclient.DefaultGasPrice,
```
chriscandless: We are the client, and we want to do some commands in a transaction. Why do we specify gas price? Shouldn't the chain define it for us?

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
+		bcs.MustMarshal(allowance),
+		l2GasBudget,
+	)
+	pt := ptb.Finish()
+
+	txData := iotago.NewProgrammable(
+		sender.Address().AsIotaAddress(),
+		pt,
+		[]*iotago.ObjectRef{},
+		2*iotaclient.DefaultGasBudget,
```
chriscandless: 1. Why it's always `2*iotaclient.DefaultGasBudget`? What is even `iotaclient.DefaultGasBudget`? Is it like the budget, which should be enough for most cases? Then why always x2?
2. If we would want to not use constant here, how would we do it? Should we pass 0 here, then do dry run?

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
-	feeCharged, err := strconv.ParseUint(estimatedReceipt.GasFeeCharged, 10, 64)
-	require.NoError(t, err)
+	// Create and execute same transaction, but with proper L2 gas budget
+	ptb = iscmoveclient.PTBCreateAndSendRequest(
+		assetsBagPTB.Clone(),
+		l1starter.ISCPackageID(),
+		env.Chain.ChainID.AsObjectID(),
+		argAssetsBag,
+		msg,
+		bcs.MustMarshal(allowance),
```
chriscandless: Why bcs here? Is it just for simplicity of passing to avoid redundant packing-unpacking?

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
 	require.NoError(t, err)
 
-	feeCharged, err := strconv.ParseUint(estimatedReceipt.GasFeeCharged, 10, 64)
-	require.NoError(t, err)
+	// Create and execute same transaction, but with proper L2 gas budget
+	ptb = iscmoveclient.PTBCreateAndSendRequest(
+		assetsBagPTB.Clone(),
+		l1starter.ISCPackageID(),
+		env.Chain.ChainID.AsObjectID(),
+		argAssetsBag,
```
chriscandless: Why there is both allowance and asset bag? Could assent bag already be used as allowance?

chriscandless (commented):
Just adding some dumb questions to boost my understanding of the code

tools/cluster/tests/estimategas_test.go
howjmay: AssetBag is the assets attached in the request. Allowance is the one that tells how many assets this call can used, which includes the assets already on chain. So allowance can be greater than AssetsBag

tools/cluster/tests/estimategas_test.go
howjmay: L1 DryRun is mostly to ensure that the format of the OnLedgerRequest is valid. Not too much logic happen there. 
The really essential one is the DryRun on L2

tools/cluster/tests/estimategas_test.go
howjmay: I think the Budget can't be 0 otherwise L1 will consider it as invalid (not enough gas to run this L1 request).
And the `DefaultGasBudget` is only a value from try and error.

tools/cluster/tests/estimategas_test.go
howjmay: If the request is a onleger request, then I think the client has to assign gas price by their own, because some of the clients may want L1 to process their request earlier. So they will give some tips, which is included in gas price

tools/cluster/tests/estimategas_test.go
howjmay: I think it is for the simplicity of the interface. @dessaya may you answer this

tools/cluster/tests/estimategas_test.go
howjmay: This is the gas budget for request on L2. It is the one for wasp nodes (EVM)

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,168 @@ package tests
...
+		msg,
+		bcs.MustMarshal(allowance),
+		l2GasBudget,
+	)
+	pt := ptb.Finish()
+
+	txData := iotago.NewProgrammable(
+		sender.Address().AsIotaAddress(),
+		pt,
+		[]*iotago.ObjectRef{},
```
howjmay: Maybe provide gas objet here. Therefore we can directly use the dry run txBytes to execute

tools/cluster/tests/estimategas_test.go
dessaya: The sender can put whatever they want in the allowance, including assets that they don't directly own on L1 (they should own them on L2 though).
So we had to settle on some format for specifying the allowance. BCS was the easiest.

tools/cluster/tests/estimategas_test.go
chriscandless: Why the test works if we pass 0?

tools/cluster/tests/estimategas_test.go
chriscandless: Oooh, I knew about that term, but was not aware we support that. Thanks!

tools/cluster/tests/estimategas_test.go
chriscandless: Nice! Fixed this

tools/cluster/tests/estimategas_test.go
howjmay: I am not sure. I thought it may because of dry run

clients/iota-go/iotaclient/gas.go
```diff
@@ -3,4 +3,5 @@ package iotaclient
 const (
 	DefaultGasBudget = 10_000_000
 	DefaultGasPrice  = 1000
+	MinGasBudget     = 1_000_000
```
chriscandless, line 6: As per https://docs.iota.org/about-iota/tokenomics/gas-in-iota#gas-budgets :
> Importantly, the minimum gas budget is 1,000,000 NANOS or 0.001 IOTA (considering that the smallest bucket maps into 1,000 computation units and assuming that reference_gas_price is 1,000 NANOS).

packages/webapi/controllers/chain/estimategas.go
```diff
@@ -16,43 +20,68 @@ import (
...
+	l1GasBudget := totalL1Gas
+	if l1GasBudget.Cmp(l1GasSummary.ComputationCost.Int) < 0 {
+		// L1 gas budget must be >= max(computation cost, total cost)
+		// See: https://docs.iota.org/about-iota/tokenomics/gas-in-iota#gas-budgets
+		l1GasBudget.Set(l1GasSummary.ComputationCost.Int)
+	}
+	if l1GasBudget.Cmp(big.NewInt(iotaclient.MinGasBudget)) < 0 {
+		// L1 gas budget must be at least 1,000,000
+		l1GasBudget.SetInt64(iotaclient.MinGasBudget)
+	}
```
chriscandless: Note that L1 gas budget is different from total gas used.

packages/webapi/controllers/chain/estimategas.go
```diff
@@ -16,43 +20,68 @@ import (
...
+	fmt.Printf("RequestBytes: %s\n", hexutil.Encode(rec.Request))
+	fmt.Printf("Request data: %v %v", res, res.Message())
+
+	// Total L1 gas = computation cost + storage cost - storage rebate
+	l1GasSummary := &dryRunResponse.Effects.Data.V1.GasUsed
+	var totalL1Gas big.Int
+	totalL1Gas.Add(&totalL1Gas, l1GasSummary.ComputationCost.Int)
+	totalL1Gas.Add(&totalL1Gas, l1GasSummary.StorageCost.Int)
+	totalL1Gas.Sub(&totalL1Gas, l1GasSummary.StorageRebate.Int)
+	// TODO: What is l1GasSummary.NonRefundableStorageFee ? Should it be included?
```
chriscandless: I dont know what this is and I was unable to find info about it at https://docs.iota.org/about-iota/tokenomics/gas-in-iota

packages/webapi/models/vm.go
```diff
@@ -46,3 +47,20 @@ func MapReceiptResponse(receipt *isc.Receipt) *ReceiptResponse {
...
+	return &L1EstimationResult{
+		GasFeeCharged: gasFeeCharged.String(),
+		GasBudget:     minGasBudget.String(),
+	}
+}
+
+type L1EstimationResult struct {
+	GasFeeCharged string `json:"gasFeeCharged,omitempty" swagger:"required,desc(Total gas fee charged (uint64 as string))"`
+	GasBudget     string `json:"gasBudget,omitempty" swagger:"required,desc(Gas budget required for processing of transaction (uint64 as string))"`
+}
```
chriscandless: Returning both fields, becaus estimation could be used for different purposes:
* For showing user how much fee will be spent (`GasFeeCharged`)
* For calculating gas budget for request (`GasBudget`)

Maybe this is not the best style. It might be confusing - the engineer will see `GasFeeCharged` and use it as gas value. Especially because we have only `GasFeeCharged` in structure for L2 part of the estimation.
So if you can come up with better style - I'm open for comments :)

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,202 @@ package tests
...
+
+		// Find proper coin objects to pay for gas
+		coinsForGas, err := env.Clu.L1Client().FindCoinsForGasPayment(context.Background(), sender.Address().AsIotaAddress(), pt, iotaclient.DefaultGasPrice, l1GasBudget)
+		require.NoError(t, err)
+
+		txData := iotago.NewProgrammable(
+			sender.Address().AsIotaAddress(),
+			pt,
+			coinsForGas,
+			2*iotaclient.DefaultGasBudget, // TODO: Why l1GasBudget here fails test?
```
chriscandless: This bothers me. I'm still trying to solve it, because without this the entire L1 estimation makes no sense.

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,202 @@ package tests
...
+	// res, err := executeTx(txBytesWithWrongL1GasBudget)
+	// require.Error(t, err)
+
+	// TODO: Should we make an attempt to execute tx with wrong L2 gas budget?
+	//       How then will it work in relation to next attempt? The request will stuck and then unstuck?
+	//
+	// // Checking that transaction execution fails with wrong L2 gas budget
+	// txBytesWithWrongL2GasBudget := createTx(l1GasBudget, l2GasBudget-1)
+	// res = executeTx(txBytesWithWrongL2GasBudget)
+	// require.NotEmpty(t, res.Errors)
```
chriscandless: These are negative cases. And they all does not work for different reasons. I'm still working on implementing them.

packages/webapi/models/vm.go
```diff
@@ -46,3 +49,43 @@ func MapReceiptResponse(receipt *isc.Receipt) *ReceiptResponse {
...
+		gasBudget.SetInt64(iotaclient.MinGasBudget)
+	}
+
+	return &L1EstimationResult{
+		ComputationFee: gasSummary.ComputationCost.Int.String(),
+		StorageFee:     gasSummary.StorageCost.Int.String(),
+		StorageRebate:  gasSummary.StorageRebate.Int.String(),
+		GasFeeCharged:  totalGas.String(),
+		GasBudget:      gasBudget.String(),
+	}
```
chriscandless: IDK what is the best name for these fields.
`ComputationFee` and `StorageFee` are called `ComputationCost` and `StorageCost` in L1 API. But term "cost" is not used in L2 API. So I used "Fee".
But "fee" is also not used in L2 - we use "GasFee" or "GasFeeCharged". But `ComputationGasFee` or `ComputationGasFeeCharged` looks unnecessary long...
So I chose just "fee".

Same with `GasFeeCharged`. In L1 its "net_gas_fee". But on L2 we use `GasFeeCharged`, and `TotalGasFeeCharged` look too long. So I just went with `GasFeeCharged`.

packages/webapi/models/vm.go
```diff
@@ -46,3 +49,43 @@ func MapReceiptResponse(receipt *isc.Receipt) *ReceiptResponse {
...
+}
+
+type L1EstimationResult struct {
+	ComputationFee string `json:"computationFee,omitempty" swagger:"required,desc(Gas cost for computation (uint64 as string))"`
+	StorageFee     string `json:"storageFee,omitempty" swagger:"required,desc(Gas cost for storage (uint64 as string))"`
+	StorageRebate  string `json:"storageRebate,omitempty" swagger:"required,desc(Gas rebate for storage (uint64 as string))"`
+	GasFeeCharged  string `json:"gasFeeCharged,omitempty" swagger:"required,desc(Total gas fee charged: computation fee + storage fee - storage rebate (uint64 as string))"`
+	GasBudget      string `json:"gasBudget,omitempty" swagger:"required,desc(Gas budget required for processing of transaction: max(computation fee, total fee) (uint64 as string))"`
```
chriscandless, line 91: There is another naming problem with these fields. For L2, GasFeeCharged is the name of field, which must be used as budget. But for L1 GasFeeCharged could be lower then the budget, so I added second field GasBudget. But it might be confusing for API users - they may accidentaly use GasFeeCharged. IDK what is the best way to solve this.

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,229 @@ package tests
...
+		testcoinRef := iotaclienttest.MintCoins(
+			t,
+			env.Clu.L1Client().IotaClient(),
+			cryptolib.SignerToIotaSigner(sender),
+			coinPackageID,
+			contracts.TestcoinModuleName,
+			contracts.TestcoinTypeTag,
+			treasuryCap,
+			1*isc.Million,
+		)
```
chriscandless: I moved this piece of code into the tx creation function, because otherwise I get error regarding version of some object. But from the names of the functions it seems to be intended to be done once.

I need to understand better how they work...

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,229 @@ package tests
...
+		if _, ok := err.(*apiclient.GenericOpenAPIError); ok {
+			msg = string(err.(*apiclient.GenericOpenAPIError).Body())
+		}
+		require.NoError(t, err, msg)
+	}
+	require.Empty(t, estimatedReceipt.L2.ErrorMessage, lo.FromPtr(estimatedReceipt.L2.ErrorMessage))
+
+	// TODO: The naming is confusing: for L1 we use "GasBudget", but for L2 we use "GasBurned".
+	l1GasBudget := lo.Must(strconv.ParseUint(estimatedReceipt.L1.GasBudget, 10, 64))
+	l2GasBudget := lo.Must(strconv.ParseUint(estimatedReceipt.L2.GasBurned, 10, 64))
```
chriscandless, line 195: The naming here is confusing: for L1 we use "GasBudget", but for L2 we use "GasBurned".
But I was unable to come up with better names here. We cannot use `GasBurned` for L1, because it is NOT "gas burned", it is a gas budget need to execute tx, which could differ from actual gas burned.
Also L2 structure already has `GasBudget`, which is budget used to run tx... 

I've added description onto API method: 
```
SetDescription("Estimates gas usage for a given on-ledger ISC request. " +
	"To calculate required L1 and L2 gas budgets use values of L1.GasBudget and L2.GasBurned respectively.").
```
If you have a good idea what are the good names here, please suggest.

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,229 @@ package tests
...
+
+	recs, err := env.Clu.MultiClient().WaitUntilAllRequestsProcessed(context.Background(), env.Chain.ChainID, res, false, 10*time.Second)
+	require.NoError(t, err, recs)
+	require.Empty(t, recs[0].ErrorMessage, lo.FromPtr(recs[0].ErrorMessage))
+	require.Equal(t, recs[0].GasBurned, estimatedReceipt.L2.GasBurned)
+	require.Equal(t, recs[0].GasFeeCharged, estimatedReceipt.L2.GasFeeCharged)
+
+	// Checking that transaction execution fails with wrong L1 gas budget
+	// TODO: Why -1 is not enough to get the error?...
+	txBytesWithWrongL1GasBudget := createTx(l1GasBudget-2000000, l2GasBudget)
```
chriscandless, line 216: For some reason not -1 nor -1_000_000 will trigger error here. This is not as estimation intended to work. So I'm going to investigate this further.

tools/cluster/tests/estimategas_test.go
```diff
@@ -2,61 +2,231 @@ package tests
...
+	require.NoError(t, err, recs)
+	require.Empty(t, recs[0].ErrorMessage, lo.FromPtr(recs[0].ErrorMessage))
+	require.Equal(t, recs[0].GasBurned, estimatedReceipt.L2.GasBurned)
+	require.Equal(t, recs[0].GasFeeCharged, estimatedReceipt.L2.GasFeeCharged)
+
+	// Checking that transaction execution fails with wrong L1 gas budget
+	// NOTE: The reason we use l1GasBudget-1200000 here is because actual storage rebate is bigger than estimated
+	//       due to mock gas coins being used for estimation. So we cannot use l1GasBudget-1 - it simply won't trigger an error.
+	//       For L1 estimated budget is not strictly equal to actual needed value, it is greater or equal. So such test is hard to write.
+	txBytesWithWrongL1GasBudget := createTx(l1GasBudget-1200000, l2GasBudget)
```
chriscandless, line 247: Not really a good test case, but I coudn't come up with anything better considering how L1 budget is calculated.

clients/iscmove/isc.go
```diff
@@ -141,8 +139,8 @@ type Message struct {
 }
 
 type Assets struct {
-	Coins   CoinBalances
-	Objects ObjectCollection
+	Coins   CoinBalances     `bcs:"optional"`
+	Objects ObjectCollection `bcs:"optional"`
 }
```
chriscandless: @howjmay These `bcs:"optional"` marks will add one byte into encoding. Why is it needed? And will it be a breaking change? Do we have `Assets` or `AssetsBagWithBalances` stores somewhere in db?

cc @lmoe

packages/solo/run.go
```diff
@@ -64,6 +65,21 @@ func (ch *Chain) EstimateGas(req isc.Request) (result *vm.RequestResult) {
 	return res.RequestResults[0]
 }
 
+// EstimateGasL1 estimates total Gas Fee, which is composed of L1 gas fee (user spent on creating onledger request)
+// and L2 gas fee (wasp gas fee for proccesing request on L2)
+func (ch *Chain) EstimateGasL1(dryRunRes *iotajsonrpc.DryRunTransactionBlockResponse) (result *vm.RequestResult, err error) {
```
chriscandless: @lmoe Isn't `EstimateGasL1` strange name for something, which
> estimates total Gas Fee, which is composed of L1 gas fee and L2 gas fee

?

clients/iscmove/isc.go
howjmay: This is in `iscmove` which should be the intermediate type for decoding, in the DB we should store `isc.Assets` instead. 

`iscmove.Assets` is used to attach Assets to a OnLedgerRequest, which may contain only Coins, or Objects only. 
Considering the move function interface receiving BCS bytes, so I think may be needed

lmoe: @chriscandless This is most likely a mistake. This will change serialization and right now neither develop or main has it. I'd remove it.

packages/solo/run.go
lmoe: @chriscandless I agree. I guess the intention was to indicate that this involves the L1 as it estimates OnLedger requests. But in it's current naming it's confusing.

lmoe (approved):

clients/iscmove/isc.go
nnikolash: I removed these tags

PR #553, opened 2025-03-20
refactor l1params

dessaya:
* Remove the `L1Params` global variable
* Add `L1ParamsFetcher`, which caches the latest L1Params and refetches it if necessary, when the epoch is out of date
* There is a single `L1ParamsFetcher` instance, owned by NodeConn.


lmoe (approved):
:fire:

PR #554, opened 2025-03-22
fix: encoding for dummy evm txs 

lmoe:
# Description of change

Dummy EVM txs are differently encoded than on Rebased. To ensure compatibility for the Tx Hashes, a legacy encoding was introduced which is only active for migrated blocks. 

Also fixes the function NativeTokens()


packages/vm/core/coreprocessors/all.go
```diff
@@ -47,5 +46,13 @@ func NewConfig() *processors.Config {
...
-	return processors.NewConfig().WithCoreContracts(lo.MergeMaps(All, Test))
+	combined := make(map[isc.Hname]isc.VMProcessor, len(All)+len(Test))
+	for k, v := range All {
+		combined[k] = v
+	}
+	for k, v := range Test {
+		combined[k] = v
+	}
+	
+	return processors.NewConfig().WithCoreContracts(combined)
```
dessaya, line 57: I'm curious, why this change?

packages/vm/core/migrations/legacy/agentid.go
```diff
@@ -0,0 +1,37 @@
...
+  - AdressKind (User address, Alias, NFT, Native) (byte)
+  - Address [32]Byte
+
+Rebased AgentID is:
+  - AgentIDKind (Ethereum, ED25519, Contract) (byte)
+  - Address [32]Byte
+
+Ethereum/Nil/Contract AgentIDs remain untouched.
+*/
+func ToLegacyAgentIDBytes(v isc.SchemaVersion, id isc.AgentID) []byte {
```
dessaya: Name is confusing IMO. It implies that it always returns the legacy version. Also since the package name is already `legacy`, it probably can be called just `legacy.AgentIDToBytes()`.

packages/vm/core/migrations/legacy/legacy_txdata_dummy_evm.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"encoding/binary"
+
+	"github.com/iotaledger/wasp/packages/isc"
+	"github.com/iotaledger/wasp/packages/kv/codec"
+	"github.com/iotaledger/wasp/packages/vm/core/migrations/allmigrations"
+)
+
+func WithDummyUint32(v isc.SchemaVersion, val uint32) []byte {
```
dessaya: Pls add some expalanation about what these functions do and why.

dessaya (approved):

packages/vm/core/coreprocessors/all.go
lmoe: lo.MergeMaps is not thread safe and multiple tests access the maps, causing a concurrent panic. I just thought I do it like so.

PR #558, opened 2025-03-25
fix: committee post tx dry run / 0 asset txs

lmoe:


clients/iscmove/iscmoveclient/iscmovetypes.go
```diff
@@ -23,10 +23,24 @@ func (ma *moveAnchor) ToAnchor() *iscmove.Anchor {
...
+// as they have to be ignored. Otherwise, the moveRequest decoding will fail.
+type intermediateMoveRequest struct {
+	ID        iotago.ObjectID
+	Sender    *cryptolib.Address
+	AssetsBag iscmove.Referent[iscmove.AssetsBag]
+	Message   iscmove.Message
+	Allowance []iscmove.CoinAllowance
+	GasBudget uint64
+}
+
```
dessaya: Honestly I would declare the type inside the `parseRequestAndFetchAssetsBag` function, since it is only used there.

clients/iscmove/isc.go
```diff
@@ -77,7 +77,7 @@ type AssetsBagBalances map[iotajsonrpc.CoinType]iotajsonrpc.CoinValue
 
 type AssetsBagWithBalances struct {
 	AssetsBag
-	Balances AssetsBagBalances `bcs:"-"`
+	Balances AssetsBagBalances `bcs:"optional"`
```
dessaya: Why is it optional? Maybe explain in a comment?

packages/nodeconn/chain.go
```diff
@@ -93,6 +93,24 @@ func (ncc *ncChain) postTxLoop(ctx context.Context) {
 		if err != nil {
 			return nil, err
 		}
+
+		dryRes, err := ncc.nodeConn.httpClient.DryRunTransaction(task.ctx, txBytes)
```
dessaya, line 99: Why is dry-run needed? The result is only used for logging? Maybe explain in a comment :pray:

dessaya (approved):

packages/nodeconn/chain.go
lmoe: DryRun runs the transaction without persisting it. 

If we ever face an issue where our Anchor progress transaction is faulty - like a MoveCall incorrectly set - it will get executed, fail and it will cost Gas. The nodes will then retry and it will do so until the Gascoin is drained.

DryRun before makes sure that the TX is correct, as we can check the Effects for success. And only afterwards we send the TX away. 

I'll add it to the comments. :)

clients/iscmove/isc.go
lmoe: Only to save space. With that you are at least free to decide if you want it or not.

packages/nodeconn/chain.go
dessaya: :+1: 

```suggestion
                // Dry-run first to avoid spending gas if the tx fails
		dryRes, err := ncc.nodeConn.httpClient.DryRunTransaction(task.ctx, txBytes)
```

clients/iscmove/isc.go
dessaya: Seems error prone. Why not make it mandatory, and if you want to save space just encode the AssetsBag?

lmoe: fine

PR #560, opened 2025-03-26
remove NFT stuff

dessaya:


lmoe (approved):

clients/iota-go/iotaclient/api_coin_query.go
```diff
@@ -29,7 +29,7 @@ func (c *Client) GetAllCoins(ctx context.Context, req GetAllCoinsRequest) (*iota
 
 type GetBalanceRequest struct {
 	Owner    *iotago.Address
-	CoinType iotago.ObjectType // optional
```
lmoe, line 32: Why was this ObjectType removed? 

Wouldn't it make sense to keep a specific type for CoinTypes coming from the API, just to indicate it properly?

clients/iota-go/iotajsonrpc/balance.go
```diff
@@ -32,6 +32,16 @@ func (t CoinType) String() string {
 	return string(t)
 }
 
+func (t *CoinType) MarshalBCS(e *bcs.Encoder) error {
+	rt, err := iotago.NewResourceType(t.String())
+	if err != nil {
+		return err
+	}
+	// use ShortString to save space
+	e.WriteString(rt.ShortString())
```
lmoe, line 41: In the beginning of our first Alpha we had an issue that we stored CoinTypes as ShortString but compared them with FullName Type strings, which caused several problems in the accounting contract. I'm not sure what the actual solution was. If we do a length agnostic check, or if we always use short/full length strings.

Just want to point out that we should validate this carefully, so we don't sometimes store full and short length types, or compare them incorrectly.

clients/iscmove/iscmoveclient/client_assets_bag.go
```diff
@@ -25,33 +25,53 @@ func (c *Client) GetAssetsBagWithBalances(
...
 		},
-		Balances: make(iscmove.AssetsBagBalances),
+		Assets: *iscmove.NewEmptyAssets(),
 	}
 	for _, data := range fields.Data {
-		resGetObject, err := c.GetObject(ctx, iotaclient.GetObjectRequest{
-			ObjectID: &data.ObjectID,
-			Options:  &iotajsonrpc.IotaObjectDataOptions{ShowContent: true},
-		})
+		// for coins the "field name" is of type 0x1::ascii::String
```
lmoe, line 31: Is this really the case for only coins and nothing else?

lmoe (approved):

clients/iota-go/iotaclient/api_coin_query.go
dessaya: I replaced the old type `ObjectType = string` with a more strict `ObjectType struct {s string}` to avoid subtle bugs where the string contains an invalid or non-canonical format.

Here I just replaced the argument type with `string`, so it keeps the functionality. I didn't remove it.

clients/iota-go/iotajsonrpc/balance.go
dessaya: Yes, that's why I defined other related types as `struct {s string}` instead of just `string`. For now let's keep it like this (if the tests pass) and when we get time to refactor properly let's unify the types and make sure this bug cannot happen.

clients/iscmove/iscmoveclient/client_assets_bag.go
dessaya: Yes, the key of the coins in the assets bag is the cointype converted to string, and for objects the key is the object ID.

:man_shrugging:

clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -64,7 +68,7 @@ func PTBTakeAndTransferCoinBalance(
 	argAssets := iotago.Argument{NestedResult: &iotago.NestedResult{Cmd: *argBorrow.Result, Result: 0}}
 	argB := iotago.Argument{NestedResult: &iotago.NestedResult{Cmd: *argBorrow.Result, Result: 1}}
 
-	for coinType, coinBalance := range assets.Coins {
+	for _, coinType := range slices.Sorted(slices.Values(maps.Keys(assets.Coins))) {
```
howjmay, line 71: why do we need to sort the coins here?

clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -262,6 +295,35 @@ func PTBReceiveRequestsAndTransition(
...
+			},
+		) {
+			obj := ptb.Command(
+				iotago.Command{
+					MoveCall: &iotago.ProgrammableMoveCall{
+						Package:       &packageID,
+						Module:        iscmove.AssetsBagModuleName,
+						Function:      "take_asset",
+						TypeArguments: []iotago.TypeTag{assetsBag.Objects[id].TypeTag()},
+						Arguments:     []iotago.Argument{argAssetsBag, ptb.MustPure(id)},
```
howjmay, line 311: If the object is deposit and withdraw in the same PTB, then ` ptb.MustPure(id)` may cause error, because the same `id` will show up more than once.

clients/iscmove/iscmoveclient/client_anchor_ptb.go
dessaya: It's to make sure that the resulting PTB is deterministic.
Probably it's not needed, but I'm not sure.

clients/iscmove/iscmoveclient/client_anchor_ptb.go
dessaya: Ok, should I replace it with ptb.MustForceSeparatePure ?

PR #562, opened 2025-03-26
Config param to disable websocket rate limit

cwarnerdev:
Adds a config parameter to disable the websocket rate limits.


lmoe (approved):
Looks good!

Sorry for being pedantic, but could you invert the configuration parameter into an Enable flag instead of Deactivate?

Most of our config parameters are like:

nodeconn.Enable
metrics.Enable

I think `websocketRateLimitEnabled` would suit better. Rather make it default to `true` inside the webapi/params.go.

PR #564, opened 2025-04-03
chore: base58 to base64

lmoe:
Based on new requirements, changing base58 to base64 for BCS fields.
An additional fix for chain deployments.

### DIFF
--- a/clients/iota-go/iotajsonrpc/dynamic_field.go
+++ b/clients/iota-go/iotajsonrpc/dynamic_field.go
@@ -8,7 +8,7 @@ import (
 // in iotago/crates/iotago-types/src/dynamic_field.rs
 type DynamicFieldInfo struct {
 	Name       iotago.DynamicFieldName                        `json:"name"`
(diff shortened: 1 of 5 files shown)


clients/iota-go/iotajsonrpc/dynamic_field.go
```diff
@@ -8,7 +8,7 @@ import (
 // in iotago/crates/iotago-types/src/dynamic_field.rs
 type DynamicFieldInfo struct {
 	Name       iotago.DynamicFieldName                        `json:"name"`
-	BcsName    iotago.Base58                                  `json:"bcsName"`
```
dessaya, line 11: Can the type Base58 be removed if it's unused?

dessaya (approved):

clients/iota-go/iotajsonrpc/dynamic_field.go
lmoe: Its still required for the Digest. That's the only last thing that's base58

PR #568, opened 2025-04-04
fix(txbuilder): send assets after consuming requests

dessaya:


clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -225,12 +225,22 @@ func PTBTakeAndPlaceToAssetsBag(
...
+	Target iotago.Address
+	Assets iscmove.Assets
+}
+
 func PTBReceiveRequestsAndTransition(
 	ptb *iotago.ProgrammableTransactionBuilder,
 	packageID iotago.PackageID,
 	argAnchor iotago.Argument,
-	requestRefs []iotago.ObjectRef,
-	requestAssets []*iscmove.AssetsBagWithBalances,
```
lmoe, line 233: We had the issue that the isc.Assets don't hold 0 base token, which is why I introduced the inclusion of the AssetsBagWithBalance. I can't right now see if this reverts my change. Does it?

lmoe (approved):

clients/iscmove/iscmoveclient/client_anchor_ptb.go
dessaya: The ConsumedRequest object still contains the AssetsBagWithBalances, so I don't think there is a change in that regard.

dessaya: But maybe I'm missing something. Was there a test for that edge case?

lmoe: Yes, a single one:

https://github.com/iotaledger/isc-rebased/pull/558/files#diff-12be2b5ad3e6179512722b78760906ac59fa8f20667fcc9761e8843487bf341cR167

dessaya: That test seems to be passing, so I guess this is fine

PR #572, opened 2025-04-09
fix: register-erc20-native-token, add: set-coin-metadata

lmoe:
This fixes 

* `register-erc20-native-token`
* depositing of native tokens / objects
 
 and adds `set-coin-metadata`.

It allows the registration of deposited L1 native tokens inside the EVM. Which will then show up in Metamask and the explorer.

### DIFF
--- a/tools/wasp-cli/chain/accounts.go
+++ b/tools/wasp-cli/chain/accounts.go
@@ -144,7 +144,8 @@ func initDepositCmd() *cobra.Command {
 			if strings.Contains(args[0], "|") {
 				// deposit to own agentID
 				tokens := util.ParseFungibleTokens(util.ArgsToFungibleTokensStr(args))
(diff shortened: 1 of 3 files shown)


dessaya (approved):

tools/wasp-cli/chain/accounts.go
```diff
@@ -144,7 +144,8 @@ func initDepositCmd() *cobra.Command {
 			if strings.Contains(args[0], "|") {
 				// deposit to own agentID
 				tokens := util.ParseFungibleTokens(util.ArgsToFungibleTokensStr(args))
-				allowance := isc.NewAssets(tokens.BaseTokens() / 10)
+				allowance := tokens.Clone()
+				allowance.SetBaseTokens(allowance.BaseTokens() / 10)
```
howjmay, line 148: Why do we divide 10 here?

PR #578, opened 2025-04-10
refactor: remove dead option and fix hints

howjmay:


go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/iotaledger/hive.go/objectstorage v0.0.0-20231018191940-f36386f4e2fa
 	github.com/iotaledger/hive.go/runtime v0.0.0-20250409140545-e1a365dbea74
 	github.com/iotaledger/hive.go/web v0.0.0-20250409140545-e1a365dbea74
+	github.com/iotaledger/iota.go/v3 v3.0.0-rc.3
```
dessaya: why is this dependency added?

howjmay: I think it is caused by `tools/stardust-migration`. I was running `go mod tidy`, and it is added.

dessaya (approved):

PR #584, opened 2025-04-11
fix: Panic when contractSandbox.Send contains oversized assets

howjmay:


packages/vm/errors.go
```diff
@@ -27,6 +27,7 @@ var (
 	ErrSenderUnknown             = coreerrors.Register("sender unknown").Create()
 	ErrNotEnoughTokensLeftForGas = coreerrors.Register("not enough funds left to pay for gas").Create()
 	ErrUnauthorized              = coreerrors.Register("unauthorized access").Create()
+	ErrAssetsOversize            = coreerrors.Register("the size of assets to be sent exceeds the max of a request")
```
dessaya, line 30:
```suggestion
	ErrTooManyAssets            = coreerrors.Register("too many sent assets").Create()
```

packages/vm/vmimpl/sandbox.go
```diff
@@ -67,8 +67,13 @@ func (s *contractSandbox) Request() isc.Calldata {
 	return s.reqctx.req
 }
 
+const ReqAssetsBagSizeLimit = 25 // this const should be equal to the one in request.move
+
 func (s *contractSandbox) Send(par isc.RequestParameters) {
 	s.Ctx.GasBurn(gas.BurnCodeSendL1Request, uint64(s.reqctx.numPostedOutputs))
+	if len(par.Assets.Coins)+len(par.Assets.Objects) > ReqAssetsBagSizeLimit {
+		panic(vm.ErrAssetsOversize)
+	}
```
dessaya, line 76: What if a request calls Send twice? One with 20 assets, and again with 20.

The request is exceeding the limit but the check will not trigger.

PR #588, opened 2025-04-14
audit: #67 fix "Logs JSON-RPC endpoint lacks validation for filter query parameters"

bas-io:


packages/evm/jsonrpc/evmchain.go
```diff
@@ -45,6 +45,8 @@ import (
 	"github.com/iotaledger/wasp/packages/vm/gas"
 )
 
+const maxLogsTopics = 4
```
lmoe, line 49: You recreate Go-Ethereums behavior here, right? Could you add a comment over the const to clarify this?

lmoe (approved):

packages/evm/jsonrpc/evmchain.go
bas-io: sure

PR #589, opened 2025-04-14
audit: #66 fix TransactionByBlockNumberAndIndex index out of bound

howjmay:


packages/evm/jsonrpc/evmchain.go
```diff
@@ -434,6 +434,9 @@ func (e *EVMChain) TransactionByBlockNumberAndIndex(blockNumber *big.Int, index
 		return nil, common.Hash{}, 0, err
 	}
 	txs := block.Transactions()
+	if index > uint64(len(txs)) {
```
dessaya: shouldn't it be `>=` ?

howjmay: I changed couple more places have the same issue

dessaya (approved):

PR #590, opened 2025-04-14
 audit: #64 fix: Avoid requests with invalid allowance panic nodeconn

howjmay:


packages/nodeconn/chain.go
```diff
@@ -218,7 +219,8 @@ func (ncc *ncChain) subscribeToUpdates(ctx context.Context, anchorID iotago.Obje
 				ncc.anchorHandler(&anchor, l1Params)
 			case req := <-newRequests:
 				onledgerReq, err := isc.OnLedgerFromRequest(req, cryptolib.NewAddressFromIota(&anchorID))
-				if err != nil {
+				// Invalid allowance will make the request execution failed, so the onledger request object will be consumed
+				if err != nil && !strings.Contains(err.Error(), "failed to parse allowance") {
```
lmoe, line 223: I don't think an error string comparison is the way to go, especially here.

It's error prone if the error string ever changes and also, if we "ignore" the request here, ISC will pull it forever, rejecting it forever as the Request object doesn't suddenly vanish.

Imo we should consume the request in some way but provoking an error in the VM, as the allowance is invalid. 

Additionally, according to the issue, panicking here causes subscribeToUpdates to kill the Goroutine as a whole which was part of the issue.

@dessaya ?

lmoe (changes requested):

packages/nodeconn/chain.go
howjmay: I ignore the error here, and propagate the request to VM. I thought VM can consume the request and return `vm.ErrNotEnoughFundsForAllowance` error in `checkAllowance()`.
Isn't this a runtime VM error that the user can see in the receipt?

PR #599, opened 2025-04-18
refactor: Update outdated webapi endpoints and models

howjmay:


packages/webapi/models/core_accounts.go
```diff
@@ -2,36 +2,14 @@ package models
...
-
-type AccountNFTsResponse struct {
-	NFTIDs []string `json:"nftIds" swagger:"required"`
+	Coins []isc.CoinJSON `json:"coins" swagger:"required"`
 }
 
-type AccountFoundriesResponse struct {
-	FoundrySerialNumbers []uint32 `json:"foundrySerialNumbers" swagger:"required"`
+type AccountObjectsResponse struct {
+	ObjectIDs []string `json:"objectIds" swagger:"required"`
```
lmoe: We have access to the object type. I think it would be good to add that as well.

packages/webapi/models/L1.go
```diff
@@ -32,6 +32,7 @@ type StateAnchor struct {
 	Raw           string `json:"raw" swagger:"desc(The raw data of the anchor (Hex)),required"`
 }
 
+// TODO should be removed in the future
```
lmoe, line 36: Why should it be removed?

packages/webapi/controllers/corecontracts/controller.go
```diff
@@ -66,34 +58,13 @@ func (c *Controller) addAccountContractRoutes(api echoswagger.ApiGroup, mocker i
...
-
-	//nolint:unused
-	type foundryOutputParams struct {
-		chainID      string `swagger:"required,desc(ChainID (Hex Address))"`
-		serialNumber uint32 `swagger:"required,min(1),desc(Serial Number (uint32))"`
-	}
-
-	api.GET("chain/core/accounts/foundry_output/:serialNumber", c.getFoundryOutput).
-		AddParamPathNested(foundryOutputParams{}).
+	api.GET("chain/core/accounts/objectdata/:objectID", c.getObjectData).
```
lmoe: We don't store object data, so we can not return it

packages/webapi/models/node.go
```diff
@@ -11,8 +11,9 @@ type VersionResponse struct {
...
 
 type InfoResponse struct {
-	Version    string               `json:"version" swagger:"desc(The version of the node),required"`
-	PublicKey  string               `json:"publicKey" swagger:"desc(The public key of the node (Hex)),required"`
-	PeeringURL string               `json:"peeringURL" swagger:"desc(The net id of the node),required"`
-	L1Params   *parameters.L1Params `json:"l1Params" swagger:"desc(The L1 parameters),required"`
+	Version    string `json:"version" swagger:"desc(The version of the node),required"`
+	PublicKey  string `json:"publicKey" swagger:"desc(The public key of the node (Hex)),required"`
+	PeeringURL string `json:"peeringURL" swagger:"desc(The net id of the node),required"`
+	// FIXME we should define a new type in package webapi to adpot to to openapi client
```
lmoe: Then please do so

clients/apiextensions/assets.go
```diff
@@ -1,43 +1,21 @@
...
 import (
-	"strconv"
-
 	"github.com/iotaledger/wasp/clients/apiclient"
 	"github.com/iotaledger/wasp/packages/coin"
 	"github.com/iotaledger/wasp/packages/isc"
 )
 
-// TODO: Handle Coins (other than base tokens) and Objects
+// TODO: Handle Objects
```
lmoe: Might make sense to just have one balance route, which returns everything an account has. Coins and Objects. Then I'd make sense to add objects here

lmoe (changes requested):
lgtm, please implement/remove any added comment

packages/webapi/models/L1.go
howjmay: The user who want to examine the past versions of Anchor can use `iota_tryGetPastObject`. And we already have a metric about `StateAnchor`. Not sure how much useful that a metric of StateTransaction is

clients/apiextensions/assets.go
howjmay: Should I rename this func for Balance ( Coins) only?

PR #616, opened 2025-05-01
store entropy in blocklog, for tracing

### LINKED ISSUE #639: Improve entropy tracing

dessaya:
closes #639


lmoe:
This looks good, but is a breaking change so don't merge it yet.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -30,6 +34,46 @@ type BlockInfo struct {
 	NumOffLedgerRequests  uint16
 	GasBurned             uint64     `bcs:"compact"`
 	GasFeeCharged         coin.Value `bcs:"compact"`
+	Entropy               hashing.HashValue
+}
+
+func init() {
+	bcs.AddCustomEncoder(func(e *bcs.Encoder, bi *BlockInfo) error {
```
chriscandless: First of all, this is not incorrect. However, I think there is a better way of doing it - defining method of BlockInfo (see `MarshalBCS(e *bcs.Encoder) error` through the code).
This `AddCustomEcoder` thingy was introduces to support encoding of third-party types. But it is not very developer-friendly. So where possible it is better to create methods.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -30,6 +34,46 @@ type BlockInfo struct {
...
+		e.Encode(bi.SchemaVersion)
+		e.Encode(bi.BlockIndex)
```
chriscandless: Again, this is NOT incorrect. However, the more performant way of encoding plain types is to use specific calls instead of `Encode()`. E.g, `WriteUint32` for `uint32`, `WriteString` etc.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -30,6 +34,46 @@ type BlockInfo struct {
...
+		e.Encode(bi.Timestamp)
+		e.EncodeOptional(bi.PreviousAnchor)
+		e.Encode(bi.L1Params)
+		e.Encode(bi.TotalRequests)
+		e.Encode(bi.NumSuccessfulRequests)
+		e.Encode(bi.NumOffLedgerRequests)
+		e.WriteCompactUint64(bi.GasBurned)
+		e.WriteCompactUint64(bi.GasFeeCharged.Uint64())
+		if bi.SchemaVersion >= blockInfoSchemaVersionAddedEntropy {
```
chriscandless: To be honest, I'm not happy, that the library forced you to do all of this manually just because of one field. Not cool, library, not cool...
However, I dont know how could we make it better.

One possible solution is something like this:
```
type BlockInfo struct {
   BlockIndex uint32
   ...
   Entropy hashing.HashValue `bcs:"-"` // NOTICE how this field is excluded from automatic encoding
}

func (bi *BlockInfo) MarshalBCS(e *bcs.Encoder) error {
   type blockInfo BlockInfo // This type is needed to avoid recursive call to this same MarshalBCS
   var plainBlockInfo blockInfo = *b
   e.Encode(&plainBlockInfo)

   if bi.SchemaVersion >= blockInfoSchemaVersionAddedEntropy {
      e.Encode(bi.Entropy)
   }

   return nil
}
```

This way we can continue just adding fields and they will be automatically encoded, and those which were skipped can be explicitly encoded in `MarshalBCS`.
But IDK, it might look confusing to everyone except of me 😄 So not worth it...

I think one of the best ways of fixing it is to add some method, which will be called after encoding and in which developer can do more encoding. So it's the same as what I did in snipper above, but automatically.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -30,6 +34,46 @@ type BlockInfo struct {
...
+		d.Decode(&bi.NumOffLedgerRequests)
+		bi.GasBurned = d.ReadCompactUint64()
+		bi.GasFeeCharged = coin.Value(d.ReadCompactUint64())
+		if bi.SchemaVersion >= blockInfoSchemaVersionAddedEntropy {
+			d.Decode(&bi.Entropy)
+		} else {
+			// we are missing entropy information; assign some unique hash for the given block index
+			bi.Entropy = hashing.HashData(bcs.MustMarshal(&bi.BlockIndex))
+		}
+		return d.Err()
```
chriscandless: NOT incorrect, but you can just return `nil` here - the `Err()` exists in encoder to allow early check for errors and early return. If you dont care about it you can just return `nil` and the library will check `Err()` later automatically

packages/vm/core/blocklog/blockinfo.go
dessaya: Good point, I'll make that change :+1:

packages/vm/core/blocklog/blockinfo.go
dessaya: Yeah, I also spent some time thinking about this. I ended up with this solution, which is the simplest, even if it does have some repetition.

Using `bcs:"-"` is a good idea... but yeah, it seems a bit confusing.

packages/vm/core/blocklog/blockinfo.go
```diff
@@ -28,8 +32,46 @@ type BlockInfo struct {
...
 	NumSuccessfulRequests uint16 // which didn't panic
 	NumOffLedgerRequests  uint16
-	GasBurned             uint64     `bcs:"compact"`
-	GasFeeCharged         coin.Value `bcs:"compact"`
+	GasBurned             uint64            `bcs:"compact"`
+	GasFeeCharged         coin.Value        `bcs:"compact"`
+	Entropy               hashing.HashValue // since v1
+}
+
+func (bi *BlockInfo) MarshalBCS(e *bcs.Encoder) error {
```
dessaya, line 40: @chriscandless is this better?

For some values (e.g. timestamp) I'm still using Encode/Decode. Is there a better way?

Also, can you check that this does not break compatibility with schema version 0? The old struct is this: https://github.com/iotaledger/wasp/blob/24e99c374a8ef976e14330734d1f3fe45acc7db7/packages/vm/core/blocklog/blockinfo.go#L22

chriscandless: Regarding `Encode/Decode` thats' fine. `Write*()` functions are intended only for built-in BCS types or plain Go types.

Regardin compatibility, also seems good. You do not do any type casting, just using matching functions. And also u used EncodeOptional where `bcs:"optional"` was and `WriteCompact*()` where `bcs:"compact"` was. Seems all correct.
The only additional action you do is `bi.GasFeeCharged.Uint64()`, but it does not change anything, because `coin.Value` already was based on `uint64`.

You could also write a very simple test:
Encode some dummy BlockInfo using original code, and print the bytes. Then use those bytes in the decoding test. I think it is nice thing to have - it will continue be useful after we bring more versions of BlockInfo.

chriscandless (approved):

packages/vm/core/blocklog/blockinfo.go
dessaya: Thanks! I did that test in blocklog_test.go :)

PR #622, opened 2025-05-03
changed workflow to push to docker hub

shufps:
push imaged to docker hub (`iotaledger/wasp:...`)


lmoe:
Was added directly to main and dev. Thx!

PR #624, opened 2025-05-08
Fix/gosec linter

cwarnerdev:
Work on clearing `gosec` linter gripes. 

Notable commits:
- None! Moved to new branch :partying_face:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -55,6 +55,7 @@ jobs:
 
   golangci:
     name: Lint
+    if: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/setup-go@v5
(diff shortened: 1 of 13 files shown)


dessaya:
I'm not sure about changing the behavior of the zero timestamp in the VM. It has been working like this since forever and I don't know if the fix will have any consequences.

cwarnerdev:
> I'm not sure about changing the behavior of the zero timestamp in the VM. It has been working like this since forever and I don't know if the fix will have any consequences.

Yeah that's why I called it out specifically. It does seem like it could cause problems but also seems like something that should be made correct. I'll move the fixes that might cause issues to another branch so they can be further evaluated without blocking us from merging the simple ones.

dessaya (approved):

PR #644, opened 2025-05-22
Tmp

nnikolash:


packages/isc/assets.go
```diff
@@ -185,9 +185,15 @@ type ObjectSet struct {
...
+func NewObjectSet(objs ...IotaObject) ObjectSet {
+	items := make(map[iotago.ObjectID]iotago.ObjectType, len(objs))
+
+	for _, obj := range objs {
+		items[obj.ID] = obj.Type
+	}
+
 	return ObjectSet{
-		items: make(map[iotago.ObjectID]iotago.ObjectType),
+		items: items,
```
nnikolash, line 196: @dessaya U good with this?

packages/vm/core/accounts/assets.go
```diff
@@ -0,0 +1,19 @@
...
+		Objects: isc.NewObjectSet(s.getAccountObjects(agentID)...),
+	}
+}
+
+func (s *StateReader) GetTotalAssets() *isc.Assets {
+	return &isc.Assets{
+		Coins:   s.getFungibleTokens(L2TotalsAccount),
+		Objects: isc.NewObjectSet(s.getL2TotalObjects()...),
+	}
+}
```
nnikolash, line 19: Please guys review this. Is it correct, that total of fungible tokens is taken from "total" acount while objects are gathered from actual user accounts?

dessaya, line 19: Yes, this seems correct, although fetching all coins & objects is a "dangerous" operation (for both a single account and the total account) if we do not set a limit for the size (hence the `TODO: implement pagination`).

packages/webapi/controllers/corecontracts/accounts.go
```diff
@@ -22,10 +22,10 @@ func (c *Controller) getTotalAssets(e echo.Context) error {
 		return c.handleViewCallError(err)
 	}
 
-	nt := assets.NativeTokens()
 	assetsResponse := &models.AssetsResponse{
-		BaseTokens:   assets.BaseTokens().String(),
-		NativeTokens: nt.JSON(),
+		BaseTokens: assets.BaseTokens().String(),
+		Coins:      assets.Coins.JSON(),
```
dessaya, line 27: Won't the Coins also include the base tokens?

packages/webapi/controllers/corecontracts/controller.go
```diff
@@ -41,22 +41,14 @@ func (c *Controller) addAccountContractRoutes(api echoswagger.ApiGroup, mocker i
...
-		SetOperationId("accountsGetAccountFoundries").
-		SetSummary("Get all foundries owned by an account")
+	// api.GET("chain/core/accounts/account/:agentID/foundries", c.getAccountFoundries).
+	// 	AddParamPath("", "chainID", "ChainID (Hex Address)").
+	// 	AddParamPath("", "agentID", "AgentID (Hex Address for L1 accounts, Hex for EVM)").
+	// 	AddParamQuery("", params.ParamBlockIndexOrTrieRoot, params.DescriptionBlockIndexOrTrieRoot, false).
+	// 	AddResponse(http.StatusUnauthorized, "Unauthorized (Wrong permissions, missing token)", authentication.ValidationError{}, nil).
+	// 	AddResponse(http.StatusOK, "All foundries owned by an account", mocker.Get(models.AccountFoundriesResponse{}), nil).
+	// 	SetOperationId("accountsGetAccountFoundries").
+	// 	SetSummary("Get all foundries owned by an account")
```
dessaya, line 51: Don't be afraid to remove

packages/webapi/controllers/corecontracts/accounts.go
nnikolash: Yes. That's by "design" :)
`BaseTokens` is just a convenient way of gettings base tokens balance, which was there before rebased. Other options is to remove that field or exclude base tokens from coins. Both options does not sound very nice.

packages/webapi/controllers/corecontracts/controller.go
nnikolash: I'm not. But I saw a lot of other commented code related to fountries. And as we plan to eventually support them again I thought commenting them is better option then removing.

packages/webapi/controllers/corecontracts/accounts.go
dessaya: I would remove the BaseTokens field, or maybe better, change the json format to a map instead of list: `{"coinType": "amount"}`, so that you can get the base tokens by `result.coins["0x2::iota::IOTA"]`.

dessaya (approved):

PR #648, opened 2025-05-22
Use safecast for uint16 conversions in DKG implementation

cwarnerdev:


packages/dkg/node.go
```diff
@@ -254,7 +266,7 @@ func (n *Node) GenerateDistributedKey(
 		edSharedPublic,
 		edPublicShares,
 		n.blsSuite,
-		uint16(deriveBlsThreshold(&initiatorInitMsg{peerPubs: peerPubs})), // TODO: Fix it.
+		derivedThreshold, // TODO: Fix it.
```
kape1395: Do we still need this?
```
// TODO: Fix it.
```

kape1395 (approved):

packages/dkg/node.go
cwarnerdev: I'm not sure, that TODO was put there by you a few years ago :D I'm not sure what it's referring to. If you think it's time to remove it I can do so.

PR #658, opened 2025-05-29
Regenereated apiclient and docs

chriscandless:


tools/wasp-cli/chain/accounts.go
```diff
@@ -41,10 +41,14 @@ func initBalanceCmd() *cobra.Command {
 			log.Check(err)
 
 			header := []string{"token", "amount"}
-			rows := make([][]string, len(balance.NativeTokens)+1)
+			rows := make([][]string, len(balance.Coins)+1)
 
 			rows[0] = []string{"base", balance.BaseTokens}
-			for k, v := range balance.NativeTokens {
+			for k, v := range balance.Coins {
+				if v.CoinType == coin.BaseTokenType.String() {
```
chriscandless: NOTE: Not printing base token here to preserve old behaviour and to not have double print of same value when only base token is present.

tools/wasp-cli/chain/accounts.go
```diff
@@ -71,13 +75,13 @@ func initAccountObjectsCmd() *cobra.Command {
...
-				AccountsGetAccountNFTIDs(ctx, agentID.String()).
+			assets, _, err := client.CorecontractsAPI.
+				AccountsGetAccountBalance(ctx, agentID.String()).
 				Execute() //nolint:bodyclose // false positive
 			log.Check(err)
 
-			for _, nftID := range nfts.NftIds {
-				log.Printf("%s\n", nftID)
+			for _, obj := range assets.Objects {
+				log.Printf("%s: %s\n", *obj.Type.S, obj.Id)
```
chriscandless: IDK what is the most beautiful style to choose here. And ID is `[]int32`...

tools/wasp-cli/chain/blocklog.go
```diff
@@ -67,15 +67,15 @@ func fetchBlockInfo(ctx context.Context, client *apiclient.APIClient, args []str
 
 	blockInfo, _, err := client.
 		CorecontractsAPI.
-		BlocklogGetBlockInfo(ctx, uint32(index)).
+		BlocklogGetBlockInfo(ctx, int32(index)).
```
chriscandless: Seems there were multiple other changes after last client re-generation, so I fixed them also.

tools/wasp-cli/chain/accounts.go
dessaya: Maybe better to use `iotago.IsSameResource` here? Because the coin type might be abbreviated.

Also maybe add a `coin.IsBaseToken()` function or similar.

tools/wasp-cli/chain/blocklog.go
dessaya: But why the change? The block index is unsigned :thinking:

chriscandless: Yeah, had no time yesterday to find the cause. Was also surprised, but assumed it's ok because generator just uses our code, so if something changes it must be ok 😅
I'll try to figure that out today.

tools/wasp-cli/chain/accounts.go
chriscandless: Nice, I was not aware of that.
IMHO that function is a little hidden right now. How about making `IsSameResource` into method? Like `EqualStr`? Or `Equal` with reflection for both `ResourceType` and `string`? 

Regarding `IsBaseToken()` - I don't know... It feels like SOLID problem. `CoinType` is type for all coins, but for some reason it knows about some special instance of itself? I would prefer leaving it as is.

packages/coin/coin.go
```diff
@@ -56,6 +56,10 @@ func ValueFromString(s string) (Value, error) {
 
 var BaseTokenType = MustTypeFromString(iotajsonrpc.IotaCoinType.String())
 
+func IsBaseToken(t string) (bool, error) {
+	return BaseTokenType.EqualsStr(t)
+}
```
chriscandless, line 61: @dessaya Good?

tools/wasp-cli/chain/accounts.go
chriscandless: Updated - added `EqualsStr()` **method** and `IsBaseToken()` **function**.

tools/wasp-cli/chain/blocklog.go
chriscandless: Fixed. Was just using `make apiclient` instead of `make apiclient-docker`

dessaya (approved):

packages/isc/assets.go
```diff
@@ -173,14 +173,30 @@ func (c CoinBalances) Clone() CoinBalances {
...
+	return IotaObjectJSON{
+		ID:   o.ID.ToHex(),
+		Type: o.Type.ToTypeJSON(),
+	}
+}
+
+type IotaObjectJSON struct {
+	ID   string                `json:"id" swagger:"required,desc(Hex-encoded object ID)"`
+	Type iotago.ObjectTypeJSON `json:"type" swagger:"required"`
+}
```
dessaya, line 198: Why does this type need to exist? Can't `iotago.ObjectID` and `iotago.ObjectType` have `MarshalJSON()` and `UnmarshalJSON()` to be automatically handled?

Maybe add a comment explaining this?

dessaya (approved):

packages/isc/assets.go
chriscandless: This type is needed to generate documentation based on it.
How it works right now:
* We specify which type is returned when defining web api controller
* That type is used to generate open yaml doc
* That doc is used to generate client code

So to have `string` type for `ID` field we need to have this type as part of value returned in API route.

As far as I know we also had "mapper" before, which allowed to use actual types and map them onto their JSON representation. But I dont see we use it anymore - I see these `*JSON` types. So I've also added mine :)

chriscandless: @dessaya  Maybe I'm wrong and I should use mapper?
@lmoe do we still use it?

dessaya: I don't know about that mapper thing. But it seems unidiomaic to have to repeat the type definition just for JSON marshalling.

If it's unavoidable, maybe it's better to define the JSON doppelganger type in `models`?

chriscandless: Yeah, 💯 !
I just saw type `CoinJSON` in the same package and thought this is the way we do it. But I agree this type makes more sense in `models` package.

chriscandless: I'll merge this PR because it is waiting for too long, and I need it for my other task. But today I'll also separately review this question and maybe fix this

chriscandless: BTW, I would say it might be convenient to have these types. They solve multiple tasks: not only they define API and documentation, they can also be used to easily implement JSON encoding instead of manually doing stuff inside `MarshalJSON`. And it is more solid approach, than having that "mapper" thingy which uses JSON definition of API which can implicitly get obsolete anytime.

chriscandless: Afte I close my other PR related to on-ledger estimation I'll try to re-implement these *JSON types using mapper (after fixing it if required).

PR #660, opened 2025-06-02
Feat/koanf

cwarnerdev:
Initial removal of viper


tools/wasp-cli/cli/config/config.go
```diff
@@ -8,11 +8,13 @@ import (
...
+
+	"github.com/knadh/koanf/parsers/json"
+	"github.com/knadh/koanf/providers/file"
 
 	"github.com/samber/lo"
 
-	"github.com/spf13/viper"
+	"github.com/knadh/koanf/v2"
 
 	"github.com/iotaledger/wasp/clients/iota-go/iotaconn"
```
dessaya: Nit: too many empty lines here.
Whenever this happens I remove all empty lines and autoformat.

Maybe we want the linter to be more strict here? or is it just me that is bothered by this?

dessaya (approved):
Approving because I don't see nothing wrong with the change. Although looking at go.sum I don't see a big difference in the amount of dependencies. This will come later?

cwarnerdev:
> Approving because I don't see nothing wrong with the change. Although looking at go.sum I don't see a big difference in the amount of dependencies. This will come later?

Viper is still indirectly imported by our iotago package so there's not really any changes with it yet. This was just a functional change. Cleaning up iotago is next!

cwarnerdev:
Viper is fully removed now :pray:

dessaya (approved):
LGTM

PR #661, opened 2025-06-03
Implement missing erc20 methods

bas-io:


packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
```diff
@@ -1298,3 +1299,93 @@ func BenchmarkRPCEstimateGas(b *testing.B) {
...
+	env := newSoloTestEnv(t)
+	_, creatorAddress := env.soloChain.NewEthereumAccountWithL2Funds()
+
+	blockNumber, err := env.Client.BlockNumber(context.Background())
+	require.NoError(t, err)
+
+	var contractAddr common.Address
+	copy(contractAddr[:], iscmagic.AddressPrefix)
+
+	// Test decimals() method - returns 9
```
dessaya: Nit: maybe better to use `t.Run()` for each subtest?

Also :open_mouth: maybe this decimals() method can save us from the whole decimals hack? cc @lmoe

packages/vm/core/evm/evmimpl/iscmagic_sandbox_view.go
```diff
@@ -62,3 +66,39 @@ func (h *magicContractHandler) GetCoinInfo(coinType iscmagic.CoinType) iscmagic.
...
+}
+
+// handler for ISCSandbox::symbol
+func (h *magicContractHandler) Symbol() string {
+	return "IOTA"
+}
+
+// handler for ISCSandbox::decimals
+func (h *magicContractHandler) Decimals() uint8 {
+	return uint8(parameters.BaseTokenDecimals)
```
dessaya: I think this should return 18 instead of 9.

packages/vm/core/evm/evmimpl/iscmagic_sandbox_view.go
```diff
@@ -62,3 +66,39 @@ func (h *magicContractHandler) GetCoinInfo(coinType iscmagic.CoinType) iscmagic.
...
+	// - totalSupply(): 0x18160ddd
+	// - balanceOf(address): 0x70a08231
+	// - transfer(address,uint256): 0xa9059cbb
+	// - transferFrom(address,address,uint256): 0x23b872dd
+	// - approve(address,uint256): 0x095ea7b3
+	// - allowance(address,address): 0xdd62ed3e
+	// XOR result: 0x36372b07
+	erc20InterfaceID := [4]byte{0x36, 0x37, 0x2b, 0x07}
+
+	return interfaceID == erc165InterfaceID || interfaceID == erc20InterfaceID
```
dessaya: I don't think the magic contract supports the full ERC20 interface right now.
https://ethereum.org/en/developers/docs/standards/tokens/erc-20/#methods

bas-io: true

packages/vm/core/evm/evmimpl/iscmagic_sandbox_view.go
bas-io: fixed

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
bas-io: t.Run() - yes

packages/vm/core/evm/evmimpl/iscmagic_sandbox_view.go
```diff
@@ -62,3 +65,30 @@ func (h *magicContractHandler) GetCoinInfo(coinType iscmagic.CoinType) iscmagic.
...
+
+// handler for ISCSandbox::symbol
+func (h *magicContractHandler) Symbol() string {
+	return "IOTA"
+}
+
+// handler for ISCSandbox::decimals
+func (h *magicContractHandler) Decimals() uint8 {
+	return uint8(18)
+}
```
dessaya: Now I'm curious: once supportsInterface(ERC20) returns false, does Metamask still call decimals and symbol? Because these methods are part of ERC20. It feels wrong to have to implement them.

dessaya: Also: h.ctx.GetCoinInfo(coin.BaseTokenType) provides the symbol.

And there is a `const ethereumDecimals = 18` in decimals_hack.go that maybe should be exported.

bas-io: I fixed symbol and decimals

Regarding `supportsInterface(ERC20)`:
Surprisingly, I did not see metamask checking ERC20 interface. The only `supportsInterface` calls I see are:
- 0x01ffc9a780ac58cd00000000000000000000000000000000000000000000000000000000 - ERC-721 (EIP-721)
- 0x01ffc9a7d9b67a2600000000000000000000000000000000000000000000000000000000 -  ERC-721 Metadata Extension interface ID

With this fix it seems that metamask still calls the same methods, hard to say what changed on UI

packages/vm/core/evm/evmimpl/iscmagic_sandbox_view.go
```diff
@@ -62,3 +66,32 @@ func (h *magicContractHandler) GetCoinInfo(coinType iscmagic.CoinType) iscmagic.
...
+func (h *magicContractHandler) Symbol() string {
+	info, ok := h.ctx.GetCoinInfo(coin.BaseTokenType)
+	h.ctx.Requiref(ok, errUnknownCoin.Error())
+	return info.Symbol
+}
+
+// handler for ISCSandbox::decimals
+func (h *magicContractHandler) Decimals() uint8 {
+	return util.EthereumDecimals
+}
```
dessaya, line 89: Ok but really I think we need to get to the bottom of this.

Why is metamask calling these? It makes no sense to me. Maybe the true reason is because of a bug somewhere else.

How are you testing it? I'll try the same.

bas-io: > How are you testing it? I'll try the same.
Connecting metamask to the chain and randomly clicking buttons for now

But why does not it make sense to you? I mean this is important metadata about the token, probably metamask uses the symbol somewhere on UI

dessaya: How does metamask know the address of the magic contract?

Do you have custom imported tokens in your metamask configuration?

bas-io: > How does metamask know the address of the magic contract?

I don't know, want to explore it

> Do you have custom imported tokens in your metamask configuration?

no

PR #670, opened 2025-06-12
Fix trace block O(N2)

bas-io:


dessaya (approved):
FYI there are linter warnings

PR #680, opened 2025-07-07
Fix state index after rotation

bas-io:


clients/iota-go/contracts/isc/sources/anchor.move
```diff
@@ -90,7 +90,12 @@ module isc::anchor {
         (Receipt { request_id }, assets)
     }
 
-    public fun transition(self: &mut Anchor, new_state_metadata: vector<u8>, mut receipts: vector<Receipt>) {
+    fun transition_internal(
+        self: &mut Anchor,
+        new_state_metadata: vector<u8>,
+        mut receipts: vector<Receipt>,
+        increment_state: bool
```
dessaya: maybe better to receive an int instead of a bool? Like `state_index_increment: u32`

clients/iota-go/contracts/isc/sources/anchor.move
```diff
@@ -102,7 +107,17 @@ module isc::anchor {
...
+        if (increment_state) {
+            self.state_index = self.state_index + 1;
+        }
+    }
+
+    public fun transition(self: &mut Anchor, new_state_metadata: vector<u8>, mut receipts: vector<Receipt>) {
+        transition_internal(self, new_state_metadata, receipts, true)
+    }
+
+    public fun transition_v2(self: &mut Anchor, new_state_metadata: vector<u8>, mut receipts: vector<Receipt>, increment_state: bool) {
```
dessaya: Maybe add some comment explaining why we have 2 versions of this function?

clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -388,12 +389,13 @@ func PTBReceiveRequestsAndTransition(
 			MoveCall: &iotago.ProgrammableMoveCall{
 				Package:       &packageID,
 				Module:        iscmove.AnchorModuleName,
-				Function:      "transition",
+				Function:      "transition_v2",
```
dessaya: How is this supposed to work on the currently deployed chains? The call to transition_v2 will break unless the move contract is upgraded.

clients/iota-go/contracts/isc/sources/anchor.move
bas-io: but why would we want to allow clients to pass arbitrary increments? bool seems more clear and restricting

clients/iota-go/contracts/isc/sources/anchor.move
bas-io: sure

clients/iscmove/iscmoveclient/client_anchor_ptb.go
bas-io: My understanding is that contract upgrade should come first, before node upgrade
Should we take care about cases when new node is running with old package id?

clients/iota-go/contracts/isc/sources/anchor.move
dessaya: Well it's debatable, and probably just a matter of taste.
I think it would be more elegant to receive an int and avoid the if in the function body. And also more flexible if e.g. for some reason in the future we need to shift the index again by some offset.

bas-io: I'm not sure about that. I think smart contracts should be as restrictive as possible

clients/iscmove/iscmoveclient/client_anchor_ptb.go
lmoe: The packageID remains the same. The version will change, and ISC will always choose the newest version of the contract.

In my understanding we can simply upgrade the contract whenever, and at some point just deploy a new version of ISC which uses the new `transition_for_rotation` function as we keep `transition` intact.

bas-io: package Id changes after upgrade

bas-io: and newest package id will be written into upgradeCap

clients/iota-go/contracts/isc/sources/anchor.move
```diff
@@ -102,9 +106,18 @@ module isc::anchor {
...
         self.state_metadata = new_state_metadata;
+    }
+
+    public fun transition(self: &mut Anchor, new_state_metadata: vector<u8>, receipts: vector<Receipt>) {
+        transition_internal(self, new_state_metadata, receipts);
         self.state_index = self.state_index + 1;
     }
 
+    /// This function performs transition for rotation (without modifying the state index)
+    public fun transition_for_rotation(self: &mut Anchor, new_state_metadata: vector<u8>, receipts: vector<Receipt>) {
```
dessaya: If we are not incrementing the state index, then I think it doesn't make sense to receive the `receipts` parameter here. It should be empty anyway.

clients/iscmove/iscmoveclient/client_anchor_ptb.go
```diff
@@ -398,5 +412,35 @@ func PTBReceiveRequestsAndTransition(
...
 	)
+
+	return ptb
+}
+
+func PTBReceiveRequestsAndTransitionForRotation(
+	ptb *iotago.ProgrammableTransactionBuilder,
+	packageID iotago.PackageID,
+	argAnchor iotago.Argument,
+	consumedRequests []ConsumedRequest,
```
dessaya: Same here, some arguments like consumedRequests, sentAssets, etc. are unused in the rotation, so maybe the function can be simplified.

packages/vm/vmtxbuilder/txbuilder.go
```diff
@@ -115,17 +115,21 @@ func (txb *AnchorTransactionBuilder) BuildTransactionEssence(stateMetadata []byt
...
+	if isRotation {
+		ptb = iscmoveclient.PTBReceiveRequestsAndTransitionForRotation(
+			txb.ptb,
+			txb.iscPackage,
+			txb.ptb.MustObj(iotago.ObjectArg{ImmOrOwnedObject: txb.anchor.GetObjectRef()}),
+			txb.consumed,
+			txb.sent,
+			stateMetadata,
+			topUpAmount,
+		)
```
dessaya: Maybe assert that `consumed` and `sent` are empty? I don't know about `topUpAmount`, maybe that one too.

clients/iota-go/contracts/isc/sources/anchor.move
bas-io: I guess so. Was not sure about that

clients/iscmove/iscmoveclient/client_anchor_ptb.go
bas-io: Wait. Do we even need this transition call at all for rotation?

tools/cluster/tests/wasp-cli.go
```diff
@@ -41,6 +41,9 @@ func newWaspCLITest(t *testing.T, opt ...waspClusterOpts) *WaspCLITest {
 		os.RemoveAll(dir)
 	})
 
+	err = os.WriteFile(path.Join(dir, "wasp-cli.json"), []byte("{}"), 0644)
+	require.NoError(t, err)
+
```
dessaya, line 47: Why is this necessary? It seems dangarous because it might overwrite an existing config, no?

dessaya (approved):

tools/cluster/tests/wasp-cli.go
bas-io: It's within the temp directory created per test execution, so nothing can be overwritten
Without this line the config in home dir is used, which I think was misleading because test execution can mess up your settings

dessaya: Makes sense. Maybe add a comment explaining this? :pray:

bas-io: sure

PR #684, opened 2025-07-09
chore: fix incompatible changes

lmoe:
Fixes:

* Block version had to be set to 6 instead of 1
* isc.Assets has been reverted back to CoinBalances 
  * WebAPI was changed to accommodate the type, but does not support returning objects

### DIFF
--- a/packages/vm/core/blocklog/blockinfo.go
+++ b/packages/vm/core/blocklog/blockinfo.go
@@ -18,12 +18,9 @@ import (
 
 const (
 	blockInfoSchemaVersion0 = iota
-	blockInfoSchemaVersion1NOOP
-	blockInfoSchemaVersion2NOOP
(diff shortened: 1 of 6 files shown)


packages/vm/core/blocklog/blockinfo.go
```diff
@@ -18,6 +18,11 @@ import (
 
 const (
 	blockInfoSchemaVersion0 = iota
+	blockInfoSchemaVersion1NOOP
+	blockInfoSchemaVersion2NOOP
+	blockInfoSchemaVersion3NOOP
+	blockInfoSchemaVersion4NOOP
+	blockInfoSchemaVersion5NOOP
```
dessaya: maybe add a comment explaining these?

Also maybe better: `blockInfoSchemaVersionBase = 5 + iota`

dessaya (approved):

PR #690, opened 2025-07-24
disable refcounts on archive nodes

dessaya:
also: improve trie commit & prune performance when refcounts are enabled


tools/wasp-cli/util/types.go
```diff
@@ -137,6 +137,10 @@ func ValueToString(vtype string, v []byte) string {
 		addr, err := codec.Decode[*cryptolib.Address](v)
 		log.Check(err)
 		return addr.String()
+	case "assets":
```
lmoe, line 145: The only nitpick I have is that we have isc.Assets and isc.CoinBalances. You name the command `assets` but decode CoinBalances.

lmoe (approved):
neat!

tools/wasp-cli/util/types.go
dessaya: Yeah, nice catch. I guess I did it this way to make it compatible with the current version in production, and then I forgot to change it back.

PR #693, opened 2025-07-31
check trie integrity in tests & fix more edge cases

dessaya:


clients/iota-go/iotaclient/iotaclienttest/api_governance_read_test.go
```diff
@@ -18,7 +18,8 @@ func TestGetCommitteeInfo(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, epochId, committeeInfo.EpochId)
 	// just use a arbitrary big number to ensure there are enough validator
-	require.Greater(t, len(committeeInfo.Validators), 3)
+	// TODO: is this assertion needed? why not >= 1?
+	// require.Greater(t, len(committeeInfo.Validators), 3)
```
howjmay: I ran this test on my local side without this change. The one before this change passed the test

dessaya: did you do `docker pull iotaledger/iota-tools:devnet` ?

howjmay: Based on the iota implementation [`crates/iota/src/iota_commands.rs:74`](https://github.com/iotaledger/iota/blob/develop/crates/iota/src/iota_commands.rs#L74), I think we should set this to 
`require.Equal(t, len(committeeInfo.Validators), 1)`

lmoe: yup, seems like it :)

lmoe (approved):

PR #695, opened 2025-08-12
Multi-chain cleanup

nnikolash:
Changes worth to note:
* config.json: `snapshotsToLoad` array became single string `snapshotToLoad` which is just block hash (without chain ID). Now only one snapshot can be loaded
* config.json: `chains` key became `chain`
* Chain registry interface modified to not require chain ID
* Added check into chain registry to forbid setting chain record with different chain ID
* Chain ID was removed from consesus messages
* Chain ID removed from PublishTX and WaitUntilAllRequestsProcessed
* Chain ID removed from events


components/chainrunner/params.go
```diff
@@ -57,15 +57,15 @@ type ParametersStateManager struct {
 }
 
 type ParametersSnapshotManager struct {
-	SnapshotsToLoad []string `default:"" usage:"list of snapshots to load; can be either single block hash of a snapshot (if a single chain has to be configured) or list of '<chainID>:<blockHash>' to configure many chains"`
-	Period          uint32   `default:"0" usage:"how often state snapshots should be made: 1000 meaning \"every 1000th state\", 0 meaning \"making snapshots is disabled\""`
-	Delay           uint32   `default:"20" usage:"how many states should pass before snapshot is produced"`
-	LocalPath       string   `default:"waspdb/snap" usage:"the path to the snapshots folder in this node's disk"`
-	NetworkPaths    []string `default:"" usage:"the list of paths to the remote (http(s)) snapshot locations; each of listed locations must contain 'INDEX' file with list of snapshot files"`
+	SnapshotToLoad string   `default:"" usage:"block hash of a snapshot to load"`
```
chriscandless, line 60: Notice `config.json` structure has changed in this place. We would need to update all configs upon deployment.

packages/webapi/services/chain.go
```diff
@@ -46,35 +47,39 @@ func NewChainService(
...
 		return err
 	}
 
 	c.log.LogInfof("StoredChainRec %v %v", storedChainRec, err)
 
 	if storedChainRec != nil {
+		if storedChainRec.ChainID() != chainRecord.ChainID() {
+			return fmt.Errorf("cannot update chain record with different chain ID: %v != %v",
+				storedChainRec.ChainID().String(), chainRecord.ChainID().String())
+		}
```
chriscandless, line 77: Added this check to ensure chain ID cannot change. This is simillar to what we had - `Delete` registry method is not exposed and `Set` had chain ID as it's argument. Now we still do not expose `Delete`, but we not dont have chain ID argument, so need to ensure nothing gets overwritten.

components/chainrunner/params.go
```diff
@@ -74,7 +74,7 @@ var (
 
 var params = &app.ComponentParams{
 	Params: map[string]any{
-		"chains":       ParamsChains,
+		"chain":        ParamsChainRunner,
```
chriscandless, line 77: Notice this key also changed in `config.json`.

components/webapi/webapi_test.go
```diff
@@ -57,6 +55,8 @@ func TestInternalServerErrors(t *testing.T) {
 	}()
 	defer e.Shutdown(context.Background())
 
+	time.Sleep(5 * time.Second)
```
chriscandless, line 58: Should fix sporadic issues in this test

packages/chain/cmtlog/cmt_log.go
```diff
@@ -60,7 +60,6 @@ type Output = map[LogIndex]*isc.StateAnchor
 
 // Protocol implementation.
 type cmtLogImpl struct {
-	chainID                isc.ChainID            // Chain, for which this log is maintained by this committee.
```
chriscandless, line 63: Notice chain ID was removed from this and many other messages in chain/chains folder. @kape1395 could you please review this part?
And in general, could removing chain ID from consensus code cause issues? E.g. make chain vulnerable to some attacks like replay attack

packages/chainrunner/runner.go
```diff
@@ -194,41 +182,22 @@ func New(
...
-				continue
-			}
-			c.snapshotsToLoad[chainID.Key()] = blockHash
-		}
+func (c *ChainRunner) initSnapshotToLoad(config string) {
+	c.snapshotToLoad = nil
+	blockHash, err := state.BlockHashFromString(config)
+	if err != nil {
+		c.log.LogErrorf("Parsing snapshots to load: %s is not a block hash: %v", config, err)
+		return
```
chriscandless, line 235: Now there is only one line, with syntax as it was for default snapshot

packages/registry/chain_registry.go
```diff
@@ -144,6 +144,8 @@ type ChainRecordRegistryEvents struct {
 }
 
 type ChainRecordRegistryImpl struct {
+	chainID isc.ChainID // We only have one chain. Multi-chain code left for backwards compatibility.
```
chriscandless: Now registry supports only one chain record. But to avoid need to update all the files we keep implementation mostly as is, with exception that interface of the registry was updated to not require chain ID for some methods. And to implement it, we store chain ID of that one chain (inside map onChange callback).

packages/chainrunner/dist/access_manager_dist.go
```diff
@@ -283,142 +269,92 @@ type accessMgrNode struct {
 	pubKey    *cryptolib.PublicKey
 	ourLC     int
 	peerLC    int
-	accessFor *chainSet
-	serverFor *chainSet
+	hasAccess bool
+	isServer  bool
```
chriscandless, line 273: Are these good names?

packages/chainrunner/accessmanager/dist/input_chain_disabled.go
chriscandless, line 1: This and other files in this folder were not created but moved. But IDK for some reason this is not diplayed even though I used git mv

packages/chainrunner/accessmanager/dist/access_manager_dist.go
```diff
@@ -283,142 +269,92 @@ type accessMgrNode struct {
 	pubKey    *cryptolib.PublicKey
 	ourLC     int
 	peerLC    int
-	accessFor *chainSet
-	serverFor *chainSet
+	hasAccess bool
+	isServer  bool
```
chriscandless, line 273: Are these good names?

packages/chainrunner/accessmanager/dist/input_chain_disabled.go
chriscandless, line 1: This and other files in this folder were not created but moved. But IDK for some reason this is not diplayed even though I used git mv

components/webapi/webapi_test.go
howjmay: I guess the system back online this part, but should be good to add it but smaller number too

chriscandless: Yeah, I just moved this line, but sure it makes sense to also decrease the time. Will do

packages/vm/core/evm/emulator/emulator.go
```diff
@@ -57,20 +57,13 @@ type GasLimits struct {
...
-	configCache, err = lru.New[int, *params.ChainConfig](100)
-	if err != nil {
-		panic(err)
-	}
-}
+var configCache atomic.Pointer[params.ChainConfig]
 
 func getConfig(chainID int) *params.ChainConfig {
-	if c, ok := configCache.Get(chainID); ok {
+	if c := configCache.Load(); c != nil {
```
dessaya, line 63: I think it's necessary to check the value of `chainID` here. Because some solo tests may still be multi-chain.

dessaya (approved):

PR #697, opened 2025-08-14
Fix/refine tests

bas-io:
cleaning up and unskipping tests


clients/iscmove/iscmoveclient/feed_test.go
```diff
@@ -14,24 +14,28 @@ import (
...
-	iscOwner := iscmoveclienttest.NewSignerWithFunds(t, testcommon.TestSeed, 0)
-	anchorOwner := iscmoveclienttest.NewSignerWithFunds(t, testcommon.TestSeed, 1)
+	iscOwner := iscmoveclienttest.NewAlphanetSignerWithFunds(t, testcommon.TestSeed, 0)
+	anchorOwner := iscmoveclienttest.NewAlphanetSignerWithFunds(t, testcommon.TestSeed, 1)
 
-	anchor := startNewChain(t, client, anchorOwner)
+	remoteNode := l1starter.NewRemoteIotaNode(iotaconn.AlphanetEndpointURL, iotaconn.AlphanetFaucetURL, cryptolib.SignerToIotaSigner(iscOwner))
+	remoteNode.Start(context.Background())
+
+	anchor := StartNewChainWithPackageIDAndL1Client(t, client, anchorOwner, remoteNode.ISCPackageID(), remoteNode.L1Client())
```
howjmay, line 32: Why do we need a new function here?

dessaya (approved):

clients/iscmove/iscmoveclient/feed_test.go
bas-io: A lot of tests and utilities for them are hardwired to l1starter.  In this particular test I need to use alphanet because local iota does not implement websockets.
So `StartNewChainWithPackageIDAndL1Client` allows you to provide specific packageId and l1Client, overriding the default which is taken from l1starter

bas-io: Other option was to provide a separate testconfig with remote L1 configuration for this particular test, but that  would require a separate module which I thought was an overkill

howjmay (approved):

PR #699, opened 2025-08-19
Refactor trie code

dessaya:
Simplify the trie interface and implementation. No behavior change.


lmoe (approved):
Looks good to me. I additionally tested the PR against our live db. Everything seems to work as expected. (I pulled random blocks between 100 - 8 million)

PR #705, opened 2025-08-26
initial glazed integration

cwarnerdev:
This is the initial integration of [glazed](https://github.com/go-go-golems/glazed) to allow standardized output from wasp CLI. There is also table-based output. This allows us to avoid fragile string and regex based parsing of CLI output in tests. 

The CLI will respond with json if passed the `--json` flag. It has `type`, `status`, `timestamp`, and `data` fields. The content of the data field is dependent on the `type`:

```json
{
  "type": "wallet_balance",
  "status": "success",
  "timestamp": "2025-08-26T14:05:00Z",
  "data": {
    "address_index": 1,
    "address": "iota1qp9427varyc05py79ajku89xarfgkj74tpel5egr6at6cu92mn5nku27mhf",
    "balances": [
      {
        "coin_type": "0x2::iota::IOTA", 
        "total_balance": "1000"
      }
    ]
  }
}
```


tools/wasp-cli/authentication/login.go
```diff
@@ -35,7 +35,12 @@ func initSetTokenCmd() *cobra.Command {
 
 			config.SetToken(node, args[0])
 
-			fmt.Printf("Set token for %s", node)
+			authOutput := format.NewAuthSuccess(node, "manual")
+			authOutput.AuthData.Message = "Token set successfully"
+			err := format.PrintOutput(authOutput)
+			if err != nil {
+				log.Printf("Error formatting output: %v", err)
+			}
```
dessaya: Maybe better `log.Check(err)` here? Same in other places with `if err != nil`

tools/wasp-cli/format/adapter.go
```diff
@@ -0,0 +1,149 @@
...
+		BaseOutput:   NewBaseOutput(commandType, "success", nil),
+		legacyOutput: legacyOutput,
+		commandType:  commandType,
+	}
+}
+
+// NewCLIOutputAdapterWithError creates a new adapter for legacy CLI output with error status
+func NewCLIOutputAdapterWithError(legacyOutput log.CLIOutput, commandType string, errorMsg string) *CLIOutputAdapter {
+	return &CLIOutputAdapter{
+		BaseOutput: NewBaseOutput(commandType, "error", map[string]interface{}{
```
dessaya: Nit: use `any` instead of `interface{}`

dessaya (approved):

tools/wasp-cli/format/wallet_address_output.go
```diff
@@ -0,0 +1,77 @@
+package format
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// WalletAddressData represents the data structure for wallet address commands
+type WalletAddressData struct {
+	AddressIndex uint32 `json:"addressIndex"`
```
howjmay, line 10: It looks like there a mix use of addressIndex and address_index. Is it intended or?

tools/wasp-cli/wallet/info.go
```diff
@@ -49,15 +37,21 @@ func initBalanceCmd() *cobra.Command {
...
-
 			balance, err := cliclients.L1Client().GetAllBalances(context.Background(), address.AsIotaAddress())
-			log.Check(err)
-
-			log.PrintCLIOutput(&BalanceModel{
-				Address:      address.String(),
-				AddressIndex: myWallet.AddressIndex(),
-				Balance:      balance,
-			})
+			if err != nil {
```
howjmay, line 41: should we use this error message? Cuz it looks more like a connection error but not a string formatting error

tools/wasp-cli/authentication/login.go
cwarnerdev: Good catch

howjmay (approved):

PR #709, opened 2025-08-31
feat: Support more transaction types

howjmay:


packages/evm/evmutil/signer.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func Signer(chainID *big.Int) types.Signer {
-	return types.NewEIP155Signer(chainID)
+	return types.NewPragueSigner(chainID)
```
dessaya, line 20: What is the consequence of changing this? Can it affect tracing old transactions?

packages/evm/jsonrpc/types.go
```diff
@@ -40,6 +42,19 @@ type RPCTransaction struct {
 	V                *hexutil.Big    `json:"v"`
 	R                *hexutil.Big    `json:"r"`
 	S                *hexutil.Big    `json:"s"`
+
+	// Typed-transaction metadata (EIP-2718)
+	Type       hexutil.Uint64   `json:"type"` // 0x0,0x1,0x2,0x3
```
dessaya, line 47: Again, will this change affect tracing old transactions?

packages/evm/evmutil/signer.go
howjmay: It supports old transactions here is the snippet from https://github.com/ethereum/go-ethereum/blob/ca6e2d141b14e67d2ba826c6213e9e413148b6ec/core/types/transaction_signing.go#L279

```
// NewPragueSigner returns a signer that accepts
// - EIP-7702 set code transactions
// - EIP-4844 blob transactions
// - EIP-1559 dynamic fee transactions
// - EIP-2930 access list transactions,
// - EIP-155 replay protected transactions, and
// - legacy Homestead transactions.
func NewPragueSigner(chainId *big.Int) Signer {
	return newModernSigner(chainId, forks.Prague)
}
```


And in `func newModernSigner(chainID *big.Int, fork forks.Fork) Signer`
it set up the legacy signer as we are doing right now
```
// configure legacy signer
	switch {
	case fork >= forks.SpuriousDragon:
		s.legacy = NewEIP155Signer(chainID)
```

So it will use EIP155 signer for legacy tx, and other signers in the corresponding scenarios

I have change the signer of a test to get the prague signer, and it works well

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
```diff
@@ -413,7 +417,7 @@ func TestRPCCall(t *testing.T) {
 	callStore := func(n uint32) {
 		receipt := env.mustSendTransactionAndWait(types.MustSignNewTx(
 			creator,
-			types.NewEIP155Signer(big.NewInt(int64(env.ChainID))),
+			env.Signer(),
```
howjmay, line 420: Here is the change to test how the prague signer works for LegacyTx

dessaya (approved):

packages/evm/jsonrpc/types.go
howjmay: I have added test `TestSystemLegacyTxChainIDOmittedStillTraceable` and `TestTraceLegacyTxInMixedLegacyAndTypedBlock`

PR #714, opened 2025-09-04
fix: Fix resource type parsing for multi-subtypes

howjmay:


packages/vm/core/blocklog/blocklog_test.go
```diff
@@ -154,7 +154,7 @@ func TestGetEventsInternal(t *testing.T) {
 
 func TestBlockInfoMarshalling(t *testing.T) {
 	t.Run("v0", func(t *testing.T) {
-		const v0hex = "002a00000000b421501f01000000640000000000000000000000000000000100000000000000000000000000000001000000000000000000000000000000e80300000000000000000000000000009edb91da930100000000000000000000005c2605000000000000000000000000000000000000000000000000000000000000000000000000000000000000000204696f746104494f5441000004496f746104494f544104494f54410f687474703a2f2f696f74612e6f726709e0afdabfb6f592bd8a010a0008000200e807f403"
+		const v0hex = "002a00000000b421501f01000000640000000000000000000000000000000100000000000000000000000000000001000000000000000000000000000000e80300000000000000000000000000009edb91da930100000000000000000000005c2605000000000000000000000000000000000000000000000000000000000000000000000000000000000000000204696f746104494f54410004496f746104494f544104494f54410f687474703a2f2f696f74612e6f726709e0afdabfb6f592bd8a010a0008000200e807f403"
```
howjmay, line 157: We replace the two the two optional fields (SubType1, SubType1) into a single optional slice: `SubTypes []*ResourceType`.
This caused one extra byte in encoding. The extra byte is the byte to identify the length of an array

howjmay:
We need to add LegacyBlockInfo, and some other Legacy types to proceed this PR. Not sure whether it is worthy

PR #718, opened 2025-09-09
fix: Fix tools/gendoc and update config_defaults.json

howjmay:


lmoe (changes requested):
The config_defaults gets generated and is not meant to be manually changed

lmoe (approved):

PR #720, opened 2025-09-09
Removed package ID from both wasp node and wasp-cli tool configs

chriscandless:


clients/l1client.go
```diff
@@ -285,6 +286,29 @@ func (c *l1Client) DeployISCContracts(ctx context.Context, signer iotasigner.Sig
...
+		return iotago.PackageID{}, fmt.Errorf("retrieving anchor object: %w", err)
+	}
+
+	objectType, err := iotago.ObjectTypeFromString(*obj.Data.Type)
+	if err != nil {
+		return iotago.PackageID{}, fmt.Errorf("parsing anchor object type: %w", err)
+	}
+
+	addr := objectType.ResourceType().Address
+	packageID, err := iotago.PackageIDFromHex(addr.String())
```
howjmay: I think you can directly cast Address to PackageId

lmoe (approved):

clients/l1client.go
chriscandless: Fixed

PR #724, opened 2025-09-15
fix: Update docker-compose in local-setup

howjmay:


howjmay:
This PR depends on the gRPC implementation #712, because the current localnet doesn't support websocket, therefore wasp will never be initialized locally.

PR #729, opened 2025-09-18
chore: Sync all the go version

howjmay:


packages/isc/vmerror.go
```diff
@@ -90,7 +90,8 @@ func (e *VMErrorTemplate) Create(params ...VMErrorParam) *VMError {
 // VMErrorTemplate by mistake, so that we don't crash the VM because of that.
 func (e *VMErrorTemplate) Error() string {
 	// calling Sprintf so that it marks missing parameters as errors
-	return fmt.Sprintf(e.messageFormat)
```
howjmay, line 93: This won't be compilable anymore.

packages/tcrypto/bls/privatekey.go
```diff
@@ -19,11 +19,12 @@ type PrivateKey struct {
 func PrivateKeyFromBytes(b []byte) (privateKey PrivateKey, err error) {
 	buffer := bytes.NewReader(b)
 
-	if privateKey, err = PrivateKeyFromMarshalUtil(buffer); err != nil {
+	privateKey, err = PrivateKeyFromMarshalUtil(buffer)
+	if err != nil {
 		err = ierrors.Wrap(err, "failed to parse PublicKey from MarshalUtil")
```
howjmay, line 24: BTW the implementation of `ierrors.Wrap()` is 
```
func Wrap(err error, message string) error {
	return fmt.Errorf("%s: %w", message, err)
}
```
Maybe we can use `fmt.Errorf()` directly?

dessaya (approved):

PR #732, opened 2025-09-26
Fix cluster tests

bas-io:


howjmay:
Maybe we need to squash the commits, some of the naming are not very proper

bas-io:
> Maybe we need to squash the commits, some of the naming are not very proper

sure. will do

tools/cluster/tests/account_test.go
```diff
@@ -9,71 +9,14 @@ import (
...
 	"github.com/iotaledger/wasp/v2/packages/vm/core/accounts"
-	"github.com/iotaledger/wasp/v2/packages/vm/core/root"
 )
 
 // executed in cluster_test.go
-func (e *ChainEnv) testBasicAccounts(t *testing.T) {
-	e.testAccounts()
-}
-
-func TestBasicAccountsNLow(t *testing.T) {
```
howjmay, line 24: May I ask the reason to remove this one?

bas-io: I noticed that with previous implementation there was significant amound of overlap in what cluster tests are testing. Simple deposit/withdrawal was tested multiple times with all kinds of configurations (1 node, 2 nodes, 3 nodes and so on). I tried to simplify it so that tests in overall run faster, increase granularity of tests and give descriptive names. 
For this particular case I renamed it to `testOffLedgerDepositWithdrawTransfer` so that it's more descriptive and I only test default cluster configuration. I think I raised this question on a meeting before questioning the purpose of running some test for N=1, N=2, etc, I mean if we just test default (4 nodes) configuration it should cover all cases

howjmay (approved):

PR #743, opened 2025-10-10
Renamed dss into distSign

chriscandless:


howjmay:
The new file names like `packages/chain/distsign/WaspDSSInsts_N4F1.cfg` still contain DSS

packages/chain/distsign/dss_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 IOTA Stiftung
```
howjmay, line 1: So does this file

howjmay (changes requested):

packages/chain/distsign/dss_test.go
chriscandless: yeah, I specifically did not rename files. Too much effort with little of profit

lmoe (approved):

PR #746, opened 2025-10-13
Renamed DKShare into DistKeyPart, and all RegistryProvides into just Registries

chriscandless:


howjmay (approved):

howjmay:
I guess we don't aim to change the filename here too right?

chriscandless:
Closing due to mixed feedback from different teammates

PR #748, opened 2025-10-20
Renamed RegistryProvider into Registry

chriscandless:


packages/chain/consensus/cons_test.go
```diff
@@ -466,7 +466,7 @@ func newTestConsInst(
 	nodeIDs []gpa.NodeID,
 	nodeStates map[gpa.NodeID]state.Store,
 	peerIdentities []*cryptolib.KeyPair,
-	dkShareRegistryProviders []registry.DKShareRegistryProvider,
+	dkShareRegistrys []registry.DKShareRegistry,
```
dessaya: `dkShareRegistries` ?

dessaya (approved):

PR #756, opened 2025-10-31
refactor GPA: typed messages

dessaya:


packages/chain/chainmanager/chain_manager.go
```diff
@@ -325,7 +326,7 @@ func (cmi *ChainMgr) handleInputAnchorConfirmed(input *inputAnchorConfirmed) gpa
 		// >         Set LatestActiveCmt <- NIL
 		// >         Set NeedConsensus <- NIL
 		if cmi.latestActiveCommittee != nil {
-			msgs.AddAll(cmi.suspendCommittee(cmi.latestActiveCommittee))
+			msgs = slices.Concat(msgs, cmi.suspendCommittee(cmi.latestActiveCommittee))
```
chriscandless, line 328: How do you feel about these inline `msgs.AddAll(doSometing)`, `slices.Concat(msgs, stateMgr.Input(...))` etc?  
Would you agree, that this style is very common in this code and that it decreases readbility? How about changing it to something like this:
```
res := cmi.suspendCommittee(cmi.latestActiveCommittee)
outmsgs = slices.Concat(msgs, resOutMsgs)
```
Because I personally loose focus when reading code like: `msgs.AddAll(stateMgr.Input(NewInputThingyDone(inputArg)))` :D

packages/chain/node.go
```diff
@@ -662,13 +662,12 @@ func (cni *chainNodeImpl) handleMilestoneTimestamp(timestamp time.Time) {
...
 func (cni *chainNodeImpl) handleNetMessage(recv *peering.PeerMessageIn) {
-	msg, err := cni.chainMgr.UnmarshalMessage(recv.MsgData)
+	msg, err := cni.chainMgr.UnmarshalPayload(recv.MsgData)
 	if err != nil {
 		cni.log.LogWarnf("cannot parse message: %v", err)
 		return
 	}
-	msg.SetSender(cni.pubKeyAsNodeID(recv.SenderPubKey))
-	cni.sendMessages(cni.chainMgr.Message(msg))
+	cni.sendMessages(cni.chainMgr.Message(gpa.NewMessageIn(cni.pubKeyAsNodeID(recv.SenderPubKey), msg)))
```
chriscandless, line 670: I would split that into 3-4 lines :D

packages/gpa/interface.go
```diff
@@ -48,95 +48,73 @@ func (niT NodeID) ShortString() string {
...
-	Recipient() NodeID // The sender should indicate the recipient.
-	SetSender(NodeID)  // The transport later will set a validated sender for a message.
+type MessagePayload interface {
 	MsgType() MessageType
 }
 
-type BasicMessage struct {
-	sender    NodeID
-	recipient NodeID
+type TypedMessageIn[T MessagePayload] struct {
```
chriscandless, line 54: Would you like `MessageInT` more?

packages/gpa/interface.go
```diff
@@ -48,95 +48,73 @@ func (niT NodeID) ShortString() string {
...
+	return &TypedMessageOut[MessagePayload]{
+		Recipient: recipient,
+		Payload:   payload,
+	}
 }
 
-func (msg *BasicMessage) SetSender(sender NodeID) {
-	msg.sender = sender
+type (
+	MessageIn  = TypedMessageIn[MessagePayload]
```
chriscandless, line 79: While I do see benefit of what you did here, I also see, that `TypedMessageIn[SomePayload]` is not automatically convertiable to `MessageIn`. When it was interface it was convertible.
Didn't this aspect create you any problem while refacroting? Because if not - maybe not is not a real drawback.

packages/chain/chainmanager/chain_manager.go
dessaya: Personally I don't find it so distracting, but I'm also not against splitting those lines if it improves readability! Let's do it.

packages/gpa/interface.go
dessaya: I don't like either :stuck_out_tongue_closed_eyes:

packages/gpa/interface.go
dessaya: Yes, this is why I added the `AsTypedMessageIn` function, e.g.:

```
func (cmi *ChainMgr) Message(msg *gpa.MessageIn) []*gpa.MessageOut {
	switch msg.Payload.(type) {
	case *msgCommitteeLog:
		return cmi.handleMsgCommitteeLog(gpa.AsTypedMessageIn[*msgCommitteeLog](msg))
	case *msgBlockProduced:
		return cmi.handleMsgBlockProduced(gpa.AsTypedMessageIn[*msgBlockProduced](msg))
	}
	panic(fmt.Errorf("unexpected message %T: %+v", msg, msg))
}
```

I'm not too happy with the ergonomics. If only Go had better type inference :(

PR #758, opened 2025-11-04
fix: Format message in JSON format for missing part in wasp-cli

howjmay:


tools/cluster/tests/wasp-cli_test.go
```diff
@@ -373,6 +363,21 @@ func TestWaspCLIDeposit(t *testing.T) {
 	// })
 }
 
+func getL2GasFee(t *testing.T, outs []string) int64 {
+	var err error
+	var l2GasFee int64
+	re := regexp.MustCompile(`Gas fee charged:\s*(\d+)`)
+
+	for _, line := range outs {
+		matches := re.FindStringSubmatch(line)
```
lmoe: Now that our cli supports structured json outputs, that shouldnt be needed anymore

lmoe (changes requested):

tools/cluster/tests/wasp-cli_test.go
howjmay: I think not all the ### cli commands accept `--json`. I am solving it

cwarnerdev (approved):
TestEVMGasPriceMismatch fails but not due to this PR

PR #781, opened 2026-01-07
fix ACSS tests

dessaya:


packages/gpa/test_context.go
```diff
@@ -213,6 +213,7 @@ func (tc *TestContext) tryProcessMessage() {
 			}
 		}
 		if gpaMsg != nil {
+			// fmt.Printf("delivering message: %s -> %s %T %+v\n", msg.Sender.ShortString(), nid.ShortString(), gpaMsg, gpaMsg)
```
dessaya, line 216: Leaving this commented out on purpose, it is useful to inspect tests but too verbose if left uncommented.

PR #788, opened 2026-01-19
fix: tests for graphql

howjmay:


packages/testutil/peering_net_behaviour_test.go
```diff
@@ -81,7 +81,7 @@ func TestPeeringNetUnreliable(t *testing.T) {
 		}
 		avgDuration /= int64(len(durations))
 		require.Greater(t, avgDuration, int64(50))
-		require.Less(t, avgDuration, int64(100))
+		require.Less(t, avgDuration, int64(120))
```
howjmay, line 84: it takes longer time now

.github/workflows/build-test.yml
```diff
@@ -51,7 +51,13 @@ jobs:
       # Here used to be a make build step, but we separated it into another job
 
       - name: Test
-        run: make test-short
+        run: |
+          TEST_SHORT_PKGS="$(go list ./... \
+            | grep -v '/packages/vm/core/testcore' \
```
howjmay, line 56: All these triggered timeout

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
```diff
@@ -704,27 +708,24 @@ func TestRPCTraceFailedTx(t *testing.T) {
...
-	require.EqualValues(t, 0, bi.NumSuccessfulRequests)
+	// No new block should be created since the tx was rejected at RPC level
+	biAfter := env.soloChain.GetLatestBlockInfo()
+	require.EqualValues(t, biBefore.BlockIndex, biAfter.BlockIndex)
 
+	// Tracing should fail because the transaction was never submitted to the chain
 	t.Run("callTracer", func(t *testing.T) {
 		_, err := env.traceTransactionWithCallTracer(tx.Hash())
-		require.ErrorContains(t, err, "expected exactly one top-level call")
+		require.ErrorContains(t, err, "transaction not found")
```
lmoe: From what I remember, `expected exactly one top-level call` has been an annoying issue with the calltracer. Why is it suddenly "transaction not found"?

It seems like the transaction wasn't executed, therefore you get transaction not found.

clients/iotagraphql/graphql_client.go
```diff
@@ -1386,7 +1434,58 @@ func (c *GraphQLClient) SignAndExecuteTransaction(
...
+	}
+	return updated, nil
+}
+
+func (c *GraphQLClient) waitForNewerObjectRef(
+	ctx context.Context,
+	current *iotago.ObjectRef,
+	timeout time.Duration,
+) (*iotago.ObjectRef, error) {
+	ticker := time.NewTicker(200 * time.Millisecond)
```
dessaya: what is the ticker for?

.github/workflows/build-test.yml
dessaya: Let's not forget to reenable or fix them. These are important tests.

clients/iotagraphql/graphql_client.go
```diff
@@ -1386,7 +1434,58 @@ func (c *GraphQLClient) SignAndExecuteTransaction(
...
+		}
+		updated[i] = fresh
+	}
+	return updated, nil
+}
+
+func (c *GraphQLClient) waitForNewerObjectRef(
+	ctx context.Context,
+	current *iotago.ObjectRef,
+	timeout time.Duration,
```
dessaya: If we already receive a context, then maybe the timeout is redundant, because the caller could pass a `ContextWithTimeout`

clients/iotagraphql/graphql_client.go
```diff
@@ -1419,25 +1518,72 @@ func (c *GraphQLClient) MintToken(
...
-				Arguments: []iotago.Argument{
-					ptb.MustObj(iotago.ObjectArg{ImmOrOwnedObject: treasuryCap}),
-					ptb.MustForceSeparatePure(mintAmount),
-					ptb.MustForceSeparatePure(signer.Address()),
+	var err error
+	var txnBytes []byte
+	var txnResponse *IotaTransactionBlockResponse
+	var gasPayments []*iotago.ObjectRef
+
+	for i := 0; i < 5; i++ {
```
dessaya: maybe the amount of retries and the delay should be parameters, or consts

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
dessaya: +1

Also, the change in graphql should not affect the EVM tests? Unless I'm missing something.

packages/solo/solo.go
```diff
@@ -685,6 +685,43 @@ func (env *Solo) L1MintCoin(
...
+	env.WaitForCoinToBeIndexed(keyPair.Address().AsIotaAddress(), coinRef.ObjectID, coinType)
+	return coinRef
+}
+
+// WaitForCoinToBeIndexed polls until the coin is available via GetCoins with the specific coin type
+func (env *Solo) WaitForCoinToBeIndexed(owner *iotago.Address, coinID *iotago.ObjectID, coinType string) {
+	ctx, cancel := context.WithTimeout(env.ctx, 60*time.Second)
+	defer cancel()
+
+	ticker := time.NewTicker(250 * time.Millisecond)
```
dessaya: Is this "permanent"? Time delays make tests unhappy.

.github/workflows/build-test.yml
howjmay: #790 is solving the timeout issues too, so we will enable soon in the future

packages/evm/jsonrpc/jsonrpctest/jsonrpc_test.go
howjmay: fixed this. It was caused by the amount coin returned by faucet is different now. I have added a test to ensure the const we used match the real amount

clients/iotagraphql/graphql_client.go
howjmay: I use it as a retry here, because the time for the node to index the TXs is too slow now. Or I should remove it?

packages/solo/solo.go
howjmay: I think we need it until we discard grahpql in solo

clients/iotagraphql/graphql_client.go
howjmay: done

clients/iotagraphql/graphql_client.go
howjmay: done

clients/iotagraphql/graphql_client.go
dessaya: Ah ok... is polling like this the "official" way? I mean, is there something like a subscription API to be notified of the updated object ref? Just trying to avoid hardcoded delays and timeouts.

howjmay: We are trying to implement the subscription of graphql. However, I think `UpdateObjectRef()` is more like a one-off call. If we don't do retry here, we add to add it to all the tests. Not sure what is the best way to do it

clients/iotagraphql/graphql_client.go
```diff
@@ -57,7 +57,46 @@ func NewGraphQLClientWithTimeout(url string, timeout time.Duration, waitParams *
...
+		}
+		time.Sleep(retryInterval)
+	}
+
+	return fmt.Errorf("timeout waiting for faucet coins to be visible")
+}
+
+func getBalance(ctx context.Context, client *GraphQLClient, address *iotago.Address) *big.Int {
+	balance, err := client.GetBalance(ctx, GetBalanceRequest{Owner: address})
+	if err != nil || balance.TotalBalance == nil {
```
dessaya, line 95: if err != nil then maybe this should panic or return error

clients/iotagraphql/graphql_client.go
dessaya: I don't know, we should try to avoid hardcoded delays if possible. I would think there must be a proper way to subscribe to updates, but I'm not familiarized with the graphql api.

Anyway, if the retry delay is unavoidable, then at least maybe it should be a parameter or a const?

clients/iotagraphql/graphql_client.go
```diff
@@ -1417,27 +1513,75 @@ func (c *GraphQLClient) MintToken(
...
-				Arguments: []iotago.Argument{
-					ptb.MustObj(iotago.ObjectArg{ImmOrOwnedObject: treasuryCap}),
-					ptb.MustForceSeparatePure(mintAmount),
-					ptb.MustForceSeparatePure(signer.Address()),
+	var err error
+	var txnBytes []byte
+	var txnResponse *IotaTransactionBlockResponse
+	var gasPayments []*iotago.ObjectRef
+
+	for i := 0; i < maxRetries; i++ {
```
dessaya, line 1526: Having to retry operations is suspicious... is the graphql api unreliable? Is it expected to have to retry? Perhaps we are using the api wrong somehow?

Maybe this is fine as a temporary measure but IMO we should at least leave TODOs to remove all delays and retries at some point.

dessaya (approved):

clients/iotagraphql/graphql_client.go
howjmay: I added the `tickingTime` in `GraphQLClient`

clients/iotagraphql/graphql_client.go
howjmay: The problem was caused by the indexer performs really slow in the test scenario. Once we switch to Simulacrum, then the problem may be gone

PR #792, opened 2026-02-11
refactor: Update local-setup for graphql setting

howjmay:


howjmay:
pending because we haven't implement subscription on graphql

PR #796, opened 2026-02-27
Fix graphql legacy types

lmoe:
# Description of change

This PR removed the legacy JSON-RPC types, conversion and mapper functions.

It also adds an experimental fake L1 simulator, that can be plugged into all tests.


packages/testutil/l1starter/l1starter.go
```diff
@@ -78,6 +78,20 @@ func IsLocalConfigured() bool {
 	return testConfig.IsLocal
 }
 
+func IsSimulatorConfigured() bool {
+	testConfig := LoadConfig()
+	return testConfig.IsSimulator
+}
+
+// TestSimulator starts the in-memory L1 simulator.
+func TestSimulator() func() {
```
howjmay, line 87: I am curious the reason to return function here

packages/test_simulator/l1/executor.go
```diff
@@ -0,0 +1,554 @@
...
+		}
+		es.cmdResults[cmdIdx] = results
+	}
+
+	// L1 gas: the simulator does not model gas costs. Real IOTA charges
+	// computation_cost (from Move VM instruction metering) + storage_cost
+	// (per-byte for new/mutated objects) - storage_rebate. Reproducing this
+	// requires executing Move bytecode, which the simulator doesn't do.
+	// TODO: Marker for adding Gas cost calculation, but this technically requires actual execution of contracts.
+	gasCost := uint64(0)
```
howjmay, line 93: Should we set a small const value as testing gas cost?

packages/testutil/l1starter/l1starter.go
lmoe: The initial idea was for clean up, so it behaves like the other functions. We don't need cleanup but at least should remain the signature.

packages/test_simulator/l1/executor.go
lmoe: That would be the best case scenario. I spent a day to improve upon this. Without a proper Move processing calculation + Gas calculation for handling the raw PTB any change in here either breaks some tests or other kinds of test.

This can be improved probably, but thats a big can of worms.

dessaya (approved):
🎉 

Too big to do a 100% detailed review. I focused on the `test_simulator` code and it looks just fine 👏

PR #817, opened 2026-03-31
skip tests that require local node

dessaya:


clients/iota-go/iotago/programmable_transaction_builder_test.go
```diff
@@ -16,12 +16,19 @@ import (
...
 
+func TestMain(m *testing.M) {
+	l1starter.TestMain(m)
+}
+
 func TestPTBMoveCall(t *testing.T) {
-	l1starter.TestLocal()
+	if !l1starter.IsLocalConfigured() {
+		t.Skip("test requires local node configuration")
+	}
```
chriscandless: I think this should be something like `if l1starter.IsSimulator() {`. Because we would still want to run this tests against real network

chriscandless (approved):

